PR #26, opened 2022-11-24
Window location search / query string example

LiamOsler:
I have been working on a project where accessing the [location search](https://developer.mozilla.org/en-US/docs/Web/API/Location/search) (query string) property of a Shinylive application is a requirement.

I thought it might be helpful for the examples to include a demonstration of the use of ui.tags.scripts along with triple quote strings for writing multi-line JavaScript "in-situ" in an app.py file, the use of Shiny.setInputValue with the "shiny:connected" listener, then how to use these values in the python portion of the script.

### DIFF
--- a/examples/location_search/about.txt
+++ b/examples/location_search/about.txt
@@ -1,1 +1,2 @@
-Retrieve the search property of the location interface.
+Location Search Property
+Retrieve the search property of the location interface
--- a/examples/location_search/about.txt
+++ b/examples/location_search/about.txt
@@ -1,0 +1,1 @@
+Retrieve the search property of the location interface.
--- a/examples/location_search/app.py
+++ b/examples/location_search/app.py
@@ -1,0 +1,31 @@
+from shiny import App, render, ui
+
+app_ui = ui.page_fluid(
+    ui.tags.script(
+       """
+        let searchObject = {};
+        let search =  window.parent.location.search.substring(1);
+        let searchArray = search.split('&');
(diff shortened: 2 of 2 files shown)


LiamOsler:
Here is a screenshot of the example in use on the https://shinylive.io/py/examples/ page:
image

wch:
Hi, thanks for the PR!

I don't think we'll add this to the examples page, since it's kind of a hack, and there are already a lot of examples and we want to keep the page from getting too much bigger. I think it's useful knowledge, but I don't know we currently have a good place for it.

For regular Shiny (not Shinylive) apps, the way we'll want people to get the query string is with something like the `session$clientData` that we use in R: https://shiny.rstudio.com/articles/client-data.html

In Python, we'll probably implement it as `session.clientdata`. Currently, the query string is available via `input['.clientdata_url_search']()`, as in this example:

```py
from shiny import App, render, ui, Session, Inputs, Outputs

app_ui = ui.page_fluid(
    ui.input_slider("n", "N", 0, 100, 20),
    ui.output_text_verbatim("txt"),
)

def server(input: Inputs, output: Outputs, session: Session):
    @output
    @render.text
    def txt():
        return f"URL search: '{input['.clientdata_url_search']()}'"

app = App(app_ui, server, debug=True)
```

Note that this will work for regular Shiny applications, but it won't work for Shinylive, since Shinylive apps run in an iframe, so this would provide the query string for the iframe, and not of the parent window.

PR #32, opened 2023-04-08
Fix reactive function name

Polkas:
A minor update with the replacement of the wrong reactive function name with a proper one in the comments.


wch:
Thank you!

PR #42, opened 2023-06-22
chore: Add step to install yarn in Makefile

karangattu:
### WHAT?
The first time I ran the command `make all` it threw an error
```sh
yarn
make: yarn: No such file or directory
make: *** [node_modules] Error 1
```
This was because `yarn` wasn't installed on the machine and it was required for installing other packages.

### HOW?
Create a variable in the makefile that checks if `yarn` exists on the machine, if not it will install it.

### POTENTIAL OPTIMIZATION?
It can probably be optimized by only invoking it when `node_modules` target is invoked.

### HOW WAS IT TESTED?
Removed `node_modules` -> `npm uninstall -g yarn` -> Run `make node_modules`. Check log to contain
```sh
➜  shinylive git:(main) make node_modules
Yarn is not installed. Installing...
yarn
```

After `yarn` is already installed, run `make node_modules` -> Check the logs to see if yarn wasn't installed again.
```sh
make node_modules
yarn
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,12 @@ MIZANI_WHEEL=mizani-$(MIZANI_VERSION)-py3-none-any.whl
 VENV = venv
 PYBIN = $(VENV)/bin
 
+# Install Yarn if not already installed
+YARN_INSTALLED := $(shell \
+	yarn -v 2> /dev/null || \
+	(echo >&2 "Yarn is not installed. Installing..."; npm install -g yarn;) \
+)
+
 
 # Any targets that depend on $(VENV) or $(PYBIN) will cause the venv to be
 # created. To use the ven, python scripts should run with the prefix $(PYBIN),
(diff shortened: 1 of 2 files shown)


wch:
Thanks for the PR! But I think we need to do it a different way.

Normally you'd want to have the installation be a make target, with something roughly like:

```
yarn:
	npm install -g yarn
```

And then you could optionally have the `all` target also ensure that it runs the `yarn` target.

However, I think in this case we don't want to install it in the Makefile at all, because yarn is something that needs to be installed outside of the project, and we don't want running `make` to alter the user's system. It would be similar to having the Makefile install `npm` automatically for the user.


Instead of changing the Makefile, I think we should just add instructions to the readme saying how to install yarn, and also point to the Yarn installation docs (https://classic.yarnpkg.com/lang/en/docs/install) because some users might not want to install it via `npm`. For example, I installed it via homebrew.

wch (changes requested):
I think we should take a different approach; see my comment above.

wch (approved):

wch:
Thanks!

PR #50, opened 2023-08-16
Apply Greg styling suggestions to orbit; use modules

jcheng5:
UI changes:

1. Use fillable plot output for simulation plot
2. Tuck direction diagram into the bottom-right corner (and hide it on smaller screen sizes)
3. Use sidebar layout
4. Put earth, moon, planetx into accordion

Code change:

1. Use Shiny modules for earth/moon/planetx
2. Separate simulation logic from plotting logic

## Before

shinylive io_py_app_ (3)

## After

localhost_60332_
localhost_60332_ (1)

### DIFF
--- a/examples/orbit/body.py
+++ b/examples/orbit/body.py
@@ -1,0 +1,56 @@
+"""Body Shiny module
+
+A Shiny module that represents a body (i.e. planet/moon) in the simulation. This allows
+us to have multiple bodies in the simulation, each sharing similar UI and server logic,
+without having to repeat the code.
+
+Learn more about Shiny modules at: https://shiny.posit.co/py/docs/workflow-modules.html
+"""
+import astropy.units as u
+import numpy as np
+from shiny import module, reactive, ui
+from simulation import Body, spherical_to_cartesian
+
+
+@module.ui
+def body_ui(enable, mass, speed, theta, phi):
+    return ui.TagList(
+        ui.input_checkbox("enable", "Enable", enable),
+        ui.panel_conditional(
+            "input.enable",
+            ui.input_numeric(
+                "mass",
+                "Mass (10^22 kg)",
+                mass,
+            ),
+            ui.input_slider(
+                "speed",
+                "Speed (km/s)",
+                0,
+                1,
+                value=speed,
+                step=0.001,
+            ),
+            ui.input_slider("theta", "Angle (𝜃)", 0, 360, value=theta),
+            ui.input_slider("phi", "𝜙", 0, 180, value=phi),
+        ),
+    )
+
+
+@module.server
+def body_server(input, output, session, label, start_vec):
+    @reactive.Calc
+    def body_result():
+        if not input.enable():
+            return None
+
+        v = spherical_to_cartesian(input.theta(), input.phi(), input.speed())
+
+        return Body(
+            mass=input.mass() * 10e21 * u.kg,
+            x_vec=np.array(start_vec) * u.km,
+            v_vec=np.array(v) * u.km / u.s,
+            name=label,
+        )
+
+    return body_result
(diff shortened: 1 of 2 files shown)


wch:
I think this looks nice, but in some ways I think it's less usable...

- All of settings panels for the bodies start collapsed, so you have to click on one of them to even know that there are settings that can be altered -- but the whole purpose of the app is to play with those settings. I think that when the app starts with those settings visible, it invites users to play with them; when it starts with the settings hidden, I think users are likely to just skip over the app. Or they'll start by clicking the "Simulate" button and all that will happen is that the plot will look exactly the same as before.
- If both Earth and Moon are open, then on a non-big screen you need to scroll down to change the moon settings, and then scroll back up to click the Simulate button. Can the button be made sticky? (And maybe it would need to go above the sliders for duration and time step.)
- Can the Earth, Moon, and Planet X headers be made bold and maybe a bit larger? As section headings, they sort of blend in with the rest of the content in the sidebar. Or maybe the header could have a different background from the body? (Also, I think that on touch devices, it think the expanding drop-down affordance may be non-obvious.)
- Can a tab panel be put in the sidebar, similar to the old layout? That would address many of these issues.
- I'm not sure if anything can be done about this here, but the focus outline on the accordion	headers only shows the horizontal bars, which I find a bit distracting. If all four sides of the outline were visible, that would be better.
image

I realize that some of these issues may be outside the scope of this app.

examples/orbit/body.py
```diff
@@ -0,0 +1,56 @@
...
+def body_server(input, output, session, label, start_vec):
+    @reactive.Calc
+    def body_result():
+        if not input.enable():
+            return None
+
+        v = spherical_to_cartesian(input.theta(), input.phi(), input.speed())
+
+        return Body(
+            mass=input.mass() * 10e21 * u.kg,
```
wch, line 50:
```suggestion
            mass=input.mass() * 1e22 * u.kg,
```

jcheng5:
All good points. Does this help?

image

> All of settings panels for the bodies start collapsed, so you have to click on one of them to even know that there are settings that can be altered -- but the whole purpose of the app is to play with those settings. I think that when the app starts with those settings visible, it invites users to play with them; when it starts with the settings hidden, I think users are likely to just skip over the app. Or they'll start by clicking the "Simulate" button and all that will happen is that the plot will look exactly the same as before.

I have to admit I always found the settings a bit overwhelming. With the above screenshot, it starts out empty so the "Simulate" button does have an effect. (It'd be really nice to disable the button after it runs, and only re-enable once a relevant parameter has changed.)

It might also be nice to have a non-primary "Randomize body settings" button, that would give a casual user an easy way to see quite different output.

> If both Earth and Moon are open, then on a non-big screen you need to scroll down to change the moon settings, and then scroll back up to click the Simulate button. Can the button be made sticky? (And maybe it would need to go above the sliders for duration and time step.)

What do you think of the above treatment--with the button alone in the corner?

> Can the Earth, Moon, and Planet X headers be made bold and maybe a bit larger? As section headings, they sort of blend in with the rest of the content in the sidebar. Or maybe the header could have a different background from the body? (Also, I think that on touch devices, it think the expanding drop-down affordance may be non-obvious.)

I agree, but simply making them bold/larger made them look kind of dumb when in the collapsed state. Probably worth discussing with @gregswinehart next week, since I don't think this problem is specific to this app.

> Can a tab panel be put in the sidebar, similar to the old layout? That would address many of these issues.

I don't believe we have that, I think the old layout (tabbed card where a sidebar would be) was the closest we could do without more custom CSS than feels appropriate for an example.

> I'm not sure if anything can be done about this here, but the focus outline on the accordion headers only shows the horizontal bars, which I find a bit distracting. If all four sides of the outline were visible, that would be better.

Agreed. It looks like the focus ring is being drawn via `box-shadow` [here](https://github.com/twbs/bootstrap/blob/803c5de30010d731b8e1043ab352a0d710ce8906/scss/_accordion.scss#L79). Changing this to an `inset` box shadow, for accordions in a sidebar, is an improvement IMHO. (cc @gadenbuie)

PR #61, opened 2023-09-12
Look for `data-engine` attribute in parent element

### LINKED ISSUE #60: run-python-blocks.ts doesn't recognize code blocks in Reveal.js
Here's what code blocks look like in `format: html`:
```html
<pre class="shinylive-r" data-engine="r"><code>#| standalone: true
```

Here's what code blocks look like in `format: revealjs':
```html
<div class="sourceCode" id="cb1" data-engine="r"><pre class="sourceCode numberSource shinylive-r number-lines code-with-copy"><code class="sourceCode"><span id="cb1-1"><a href="#cb1-1"></a>#| standalone: true</span>
```

Relevant code: https://github.com/posit-dev/shinylive/blob/ff9d728298e1bf570418267fe1d91cc1b58dfc2e/src/run-python-blocks.ts#L13C1-L20

wch:
This closes #60.

This is the HTML structure that shinylive expects:

```html
<pre class="shinylive-r" data-engine="r"><code>#\| standalone: true
...
```

Shinylive searches for elements with `class="shinylive-r"`, and then looks at the `data-engine` attribute on that element.

However, the RevealJS format for Quarto transforms it into this:

```html
<div class="sourceCode" id="cb2" data-engine="r">
  <pre class="sourceCode numberSource shinylive-r number-lines code-with-copy">
  <code class="sourceCode">
  <span id="cb2-1"><a href="#cb2-1"></a>#\| standalone: true</span>
  ...
```

Note that the `data-engine` attribute is no longer in the same element as `class="shinylive-r"`.

This PR makes it so that it will look for `data-engine` in both the `class="shinylive-r"` element, and the immediate parent.

I think that a better fix might be to update the revealjs format so that it moves `data-` attributes to the child, along with the class.

This document now works:

	---
	title: "Shinylive RevealJS"
	format: revealjs
	filters:
	  - shinylive
	---
	
	## Shinylive Python
	
	```{shinylive-python}
	#| standalone: true
	#| viewerHeight: 420
	from shiny import App, render, ui
	
	app_ui = ui.page_fluid(
	    ui.h2("Hello Shiny!"),
	    ui.input_slider("n", "N", 0, 100, 20),
	    ui.output_text_verbatim("txt"),
	)
	
	
	def server(input, output, session):
	    @output
	    @render.text
	    def txt():
	        return f"n*2 is {input.n() * 2}"
	
	
	app = App(app_ui, server)
	```
	
	## Shinylive R
	
	```{shinylive-r}
	#| standalone: true
	#| viewerHeight: 420
	ui <- fluidPage(
	  sliderInput("obs", "Number of observations:",
	    min = 0, max = 1000, value = 500
	  ),
	  plotOutput("distPlot")
	)
	
	server <- function(input, output) {
	  output$distPlot <- renderPlot({
	    hist(rnorm(input$obs))
	  })
	}
	
	shinyApp(ui, server)
	```

### DIFF
--- a/src/run-python-blocks.ts
+++ b/src/run-python-blocks.ts
@@ -1,6 +1,6 @@
 import type { AppEngine, AppMode } from "./Components/App";
-import { parseCodeBlock } from "./parse-codeblock";
 import type { Component } from "./parse-codeblock";
+import { parseCodeBlock } from "./parse-codeblock";
 // @ts-expect-error: This import is _not_ bundled. It would be nice to be able
 // import type information from ./Components/App (which gets compiled to
 // shinylive.js), but I haven't figured out how to make it do that and do this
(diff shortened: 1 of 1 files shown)


jcheng5:
Is it even necessary to look at the separate `engine` attribute? We've already matched on `.shinylive-r` or `.shinylive-python`, could we use that to infer the engine?

wch:
That's a really good point. I'd be fine with relying on the `class` instead of `engine`.

@georgestagg, what do you think?

georgestagg:
Yes, extracting the engine from the class name instead seems reasonable to me.

I had liked the data attribute, it seemed elegant to pass the information that way. But, the reveal JS issue makes it not such an elegant method after all.

wch:
Great, I'll go ahead and just use the class, then.

PR #63, opened 2023-09-13
Add a null LSP client for use when not running Python apps

georgestagg:
This PR adds a new "null" LSP client. The LSP client starts up, but does nothing. Writing out messages is ignored and incoming messages are never generated.

The idea here is rather than conditionally executing code in many places related to setting up and using the pyright LSP, we can instead create a dummy LSP that does nothing for use when running R Shiny apps.

Later, we can add a new R LSP client class and switch to using that for R apps instead.

Such a null LSP client might also be useful in the future as part of an abstract client that delegates messages to the various LSP servers we eventually support, with the null client as a fallback to quietly handle file types where there is no LSP server.

With this change, shinylive should no longer try to read any files from `/pyright` or `/pyodide` when running with the R engine.

### DIFF
--- a/src/Components/Editor.tsx
+++ b/src/Components/Editor.tsx
@@ -6,6 +6,7 @@
 import * as fileio from "../fileio";
 import { createUri } from "../language-server/client";
 import { LSPClient } from "../language-server/lsp-client";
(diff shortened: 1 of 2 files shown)


wch (approved):
Looks great!

PR #64, opened 2023-09-13
Remove default engine argument in `runApp()`, ensure correct engine is selected at runtime

georgestagg:
In this PR I have reworked how the R and Python shinylive websites are built so that it no longer depends on using different versions of the `shinylive` asset pack.

Previously, this was handled by setting a default value for the `engine` argument for `runApp()`, with a default that depended on if `make buildjs` or `make buildjs-r` has been run. This turned out to be a Bad Idea™, due to the practice of symlinking to the latest build of the shinylive assets with `r-shinylive` and/or `py-shinylive`.

The mechanism described above is removed in this PR. Instead, the `engine` argument to `runApp()` is now always required. (Note: This could be set to default to "python" without any loss of generality, if we liked).

The HTML pages that load the Shinylive for R or Shinylive for Python website are now built from templates in `site_templates/` and written to `site/[...]/index.html` by `build.ts`. The templating interpolates the value of engine to "r" or "python", based on if e.g. either `make serve` or `make serve-r` was run.

Some component logic has been moved around to avoid the build-time defined `process.env.APP_ENGINE`, and that define  has been removed entirely.

Importantly, with this PR it no longer matters if the R or Python shinylive website was built last - the resulting asset pack is the same. This is almost certainly a good thing.

---

We now must ensure that `runApp()` is invoked with an `engine` argument everywhere. Here are the locations I am aware of:

* `src/run-python-blocks.ts` - Already always passes the correct engine thanks to #61.
* `site/[...]/index.html` - Templated using `{{APP_ENGINE}}` in this PR.
* `export_template/[edit]/index.html` - Should be templated using `{{APP_ENGINE}}` by `r-shinylive` and `py-shinylive`

---

The `r-shinylive` package needs to be modified to set the `runApp()` engine argument to `"r"`. The current version does not pass the argument at all and that won't work. I have made a PR () that does the required template replacement.

An issue is that older versions of `py-shinylive` won't know that it needs to template the `{{APP_ENGINE}}` string. In this PR a check is also added: if the interpolation has not been done the engine will default to using `python`. This should mean old versions of py-shinylive continue to work with the latest shinylive assets.

I think in the future, when `app.json` contains an `engine` property, we can remove this templating.


schloerke:
> An issue is that older versions of py-shinylive won't know that it needs to template the {{APP_ENGINE}} string. In this PR a check is also added: if the interpolation has not been done the engine will default to using python. This should mean old versions of py-shinylive continue to work with the latest shinylive assets.

py-shinylive is tied to a single version of shinylive assets. So there is no worry. I'll make a PR to py-shinylive to support `{{APP_ENGINE}}` for this release.

georgestagg:
Brill, with that I think we can simplify,

```js
      let engine = "{{APP_ENGINE}}";
      if (engine.includes("APP_ENGINE")) {
        engine = "python";
      }
```

to just,

```js
      let engine = "{{APP_ENGINE}}";
```

Makefile
```diff
@@ -233,10 +233,10 @@ serve-r:
 
 # Build the _shinylive directory for deployment of both R and Python sites
 _shinylive:
-	$(MAKE) buildjs-prod
-	cp -Lr $(SITE_DIR) $(SHINYLIVE_DIR)/py
 	$(MAKE) buildjs-prod-r
-	cp -Lr $(SITE_DIR) $(SHINYLIVE_DIR)/r
+	cp -Lr $(SITE_DIR)/. $(SHINYLIVE_DIR)/r
+	$(MAKE) buildjs-prod
+	cp -Lr $(SITE_DIR)/. $(SHINYLIVE_DIR)/py
```
wch, line 239: Just making sure: the order of these builds doesn't matter anymore, right?

georgestagg: Correct, the order change is just left over from earlier experimentation. I should flip it back to make the intention clearer here.

The addition of `/.` to the end of the path is important, though. Without it you end up with `$(SHINYLIVE_DIR)/r/r` if `$(SHINYLIVE_DIR)/r` already exists.

wch (approved):

PR #70, opened 2023-10-04
Determine engine from first arg in codeblock-to-json.ts

schloerke:
Pass in `engine` as first `arg` to `codeblock-to-json.js` so that quarto extension can tell which language is being used.


schloerke:
Reviewed `arg` retrieve with @georgestagg in person. Merging.

PR #82, opened 2023-11-28
Load R Shiny package from filesystem image

georgestagg:
As in #80.

Downloads the Shiny R package filesystem library for distribution as part of the Shinylive assets. R Shiny and its dependences are loaded from the copy in the local asset bundle. Any other R packages loaded in the shinylive app continue to be loaded from CDN.

- [x] Ensure large assets can be used with shinylive.io.


georgestagg:
Node: This PR will need to be merged as part of the prep for posit-dev/r-shinylive#63.

PR #79, opened 2023-11-16
Pass URL query parameters to the Shiny app in `Viewer.tsx`

georgestagg:
Fixes posit-dev/r-shinylive#28

I saw someone ask for this recently, the idea being that the query parameters in a shinylive.io URL should be forwarded and accessible to the inner Shiny app running in the `iframe` on the page.

I am unsure if we actually would want this behaviour or not, I could be convinced either way. But in any case, the implementation is simple (grab the parameters from the outer page URL and append to the `iframe`'s `src` attribute) so I have placed the change here for discussion.

Also includes some formatting changes from prettier, apologies for the noise.

- [ ] Add enable/disable option for passing URL parameters - default to off?
- [ ] Consider how this feature interacts with `runApp()`.

PR #84, opened 2023-11-30
Return pass through the query string and hash

schloerke:
Fixes posit-dev/r-shinylive#28

Q:
- [ ] Should we support this?
- [ ] Would the query string and hash be passed through to every app on a quarto document? If so, is that OK?


schloerke:
Closing in favor of #79

PR #96, opened 2024-01-12
Remove apps

gshotwell:
This reorganizes the shinylive site in accordance with https://github.com/posit-dev/py-shiny/issues/995. I decided to keep all the applications there and just change the JSON in case we want to add anything back.


gadenbuie (approved):
Looks good to me! FYI I think there are some overlapping commits with earlier batches

PR #99, opened 2024-01-22
Expressify shiny widgets

gshotwell:
I mostly included the  apps from the new ipywidget docs. One bit that might need a callout is the `render.express` example. I wanted something which returned a context manager, and iterator, and a rendering function, and to do that I needed to be explicit aobut the output. This is basically a module workaround, but I think it's a pretty good example of what `render.express`can do. 

```
from shiny import render
from shiny.express import ui, input, output
import numpy as np
import matplotlib.pyplot as plt


ui.input_slider("n", "Number of cards", value=3, min=1, max=5)


@render.express
def cards():
    with ui.layout_columns():
        for i in range(input.n()):
            with ui.card():
                f"Card {i}"

                @output(id=f"hist_{i}")
                @render.plot(alt="A histogram")
                def histogram():
                    np.random.seed(19680801)
                    x = 100 + 15 * np.random.randn(437)
                    plt.hist(x, 20, density=True)
```

### DIFF
--- a/examples/python/output_ui/app.py
+++ b/examples/python/output_ui/app.py
@@ -27,5 +27,4 @@ def custom_card(id):
 def cards():
     with ui.layout_columns():
         for i in range(input.card_n()):
-            with ui.card():
-                custom_card(i)
+            custom_card(i)
--- a/examples/python/output_ui/app.py
+++ b/examples/python/output_ui/app.py
@@ -1,36 +1,24 @@
 from shiny import render
-from shiny.express import ui, input
+from shiny.express import ui, input, output
+import numpy as np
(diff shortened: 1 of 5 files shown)


examples/python/output_ui/app.py
```diff
@@ -1,36 +1,30 @@
 from shiny import render
-from shiny.express import ui, input
+from shiny.express import ui, input, output, expressify
```
gadenbuie, line 2: Import `render` from `shiny.express`. We should change this now that we noticied but we'll also need to do a pass before merging any `rc-express` branches.

gadenbuie (approved):

PR #129, opened 2024-05-02
Update browser URL hash when user presses play button

### LINKED ISSUE #128: Feature Request: Update URL for current app when running
It took me a long time to learn **not to** copy the URL of the current page to share an app in editor mode because the URL will point to the initially loaded app and not the current version being edited.

It'd be helpful if the URL were updated in editor mode to reflect the current state of the app each time it is re-run via the Play button or <kbd>Cmd/Ctrl</kbd> + <kbd>Shift</kbd> + <kbd>Enter</kbd>.

wch:
This closes #128. When the user presses the play button, it will automatically update the URL hash.

The encoding is done in a web worker, so it will not block the main thread.

A few remaining tasks:
- [x] Don't auto-update the if files are very large -- need to figure out how to indicate to user in those cases that they should click the share button.
- [x] Only update the URL if shinylive is in full-page mode on the examples/ or editor/ URL. Don't update if a shinylive app is embedded in another web page, or if someone is editing an app that they've created a static deployment for.
- [x] Maybe don't auto-update if the code is loaded from a gist? Not sure what the behavior should be for this case.
- [x] Only update the URL if the code has been changed (this might not be necessary, though).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "shinylive",
-  "version": "0.2.8",
+  "version": "0.3.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "shinylive",
-      "version": "0.2.8",
+      "version": "0.3.0",
       "license": "MIT",
       "devDependencies": {
         "@codemirror/autocomplete": "^6.4.2",
@@ -61,6 +61,7 @@
         "prettier-plugin-organize-imports": "^3.2.4",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
+        "react-hot-toast": "^2.4.1",
         "tsx": "^4.7.0",
         "typescript": "^5.3.3",
         "vscode-languageserver-protocol": "^3.17.5",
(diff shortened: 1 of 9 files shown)


gadenbuie:
> * need to figure out how to indicate to user in those cases that they should click the share button.

Maybe a toast message that says something like "Auto-updating the app link is disabled. Please use the share button."?



> * Maybe don't auto-update if the code is loaded from a gist? Not sure what the behavior should be for this case.

Can `gist` and `code` coexist in the URL? I think the expectation is that copying the current URL will give you a link to the app you're editing right then. So to me it makes sense that if you open a gist, update the code and then copy the URL, that URL would restore the _edited app_.

Keeping the gist provenance seems reasonable so you could delete `code=...` to reset, but if they're mutually exclusive then I think it's okay to drop `gist` from the URL.

wch:
> > * need to figure out how to indicate to user in those cases that they should click the share button.
> 
> Maybe a toast message that says something like "Auto-updating the app link is disabled. Please use the share button."?

That sounds reasonable -- I did think about that earlier and thought it would be too intrusive, but now that I think about it again, it would only happen in rare cases anyway.


> > * Maybe don't auto-update if the code is loaded from a gist? Not sure what the behavior should be for this case.
> 
> Can `gist` and `code` coexist in the URL? I think the expectation is that copying the current URL will give you a link to the app you're editing right then. So to me it makes sense that if you open a gist, update the code and then copy the URL, that URL would restore the _edited app_.
> 
> Keeping the gist provenance seems reasonable so you could delete `code=...` to reset, but if they're mutually exclusive then I think it's okay to drop `gist` from the URL.

`gist` and `code` can't coexist, so I suppose it does make sense to replace the `gist=` with `code=` if they re-run the app. Ideally it would only be replaced if the user actually modifies the files, and I'll look into making it work that way.

PR #140, opened 2024-06-13
feat: Allow CSS units in `viewerHeight` option

gadenbuie:
* Fixes https://github.com/quarto-ext/shinylive/issues/36

Adds a utility function, `asCssLengthUnit()`, that promotes numbers to pixels or otherwise passes through a string value. This enables `viewerHeight: 300` → `300px` as in the current behavior or `viewerHeight: 300px` (or any other CSS unit).

### DIFF
--- a/src/Components/utils.ts
+++ b/src/Components/utils.ts
@@ -1,0 +1,15 @@
+export function asCssLengthUnit(value?: number | string): string | undefined {
+  if (value === undefined) {
+    return undefined;
+  }
+
+  if (typeof value === "string") {
+    return value;
+  }
+
+  if (typeof value !== "number") {
+    return undefined;
+  }
+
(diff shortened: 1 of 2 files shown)


wch (approved):

src/Components/utils.ts
```diff
@@ -0,0 +1,15 @@
...
+    return undefined;
+  }
+
+  if (typeof value === "string") {
+    return value;
+  }
+
+  if (typeof value !== "number") {
+    return undefined;
+  }
```
wch, line 12: When would it get here?

gadenbuie: Yeah, it'd have to be a value that makes it past the type checker. Happy to take this out if its unnecessary.

PR #142, opened 2024-06-13
feat: Add assignment and pipe keybindings

gadenbuie:
This PR adds two common and highly popular keybindings to the editor for the R app engine:

* <kbd>Alt/Opt</kbd> + <kbd>-</kbd>: Assignment operator `<-`
* <kbd>Ctrl/Cmd</kbd> + <kbd>Shift</kbd>  + <kbd>M</kbd>: Pipe operator `|>`

Given that r-wasm works with R >= 4.3, I think it's fair to use the base-R pipe operator and not worth the additional effort required to make the keybinding configurable.

The keybindings take care to only add a leading and trailing space when there isn't already space around the current cursor position.

https://github.com/posit-dev/shinylive/assets/5420529/36953102-9ca5-4977-a24b-6b592c08870b

### DIFF
--- a/src/Components/Editor.tsx
+++ b/src/Components/Editor.tsx
@@ -3,8 +3,8 @@
 // released; when it's released, we can remove this.
 // https://github.com/microsoft/vscode/issues/141908
 /// <reference types="wicg-file-system-access" />
-import type { Extension } from "@codemirror/state";
-import { EditorState, Prec } from "@codemirror/state";
+import type { Extension, SelectionRange } from "@codemirror/state";
+import { EditorSelection, EditorState, Prec } from "@codemirror/state";
 import type { KeyBinding, ViewUpdate } from "@codemirror/view";
 import { EditorView, keymap } from "@codemirror/view";
 import "balloon-css";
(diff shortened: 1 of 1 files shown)


wch:
Looks good overall!

A couple of issues I noticed:
- If text is selected and you use one of these shortcuts, the selected text is _not_ replaced. Instead, the selection becomes unselected, and then the text is inserted at the cursor location.
- If there are multiple cursors, then using these shortcuts only inserts the text at the first cursor, and also makes all the other cursors disappear.

gadenbuie:
> * If text is selected and you use one of these shortcuts, the selected text is _not_ replaced. Instead, the selection becomes unselected, and then the text is inserted at the cursor location.

I was okay with this, but I don't have strong feelings either way. This is how insertion works in vscode with vscode-R, so I'm pretty used to it now, but in RStudio the selection is replaced.

gadenbuie:
@wch Okay, I've fixed both of the issue you pointed out. The keybinding now replaces the current selection and supports multicursor insertion. It also works out correct whitespace for each cursor.

https://github.com/posit-dev/shinylive/assets/5420529/6ebda397-2b3c-45b1-af45-f4e204d023c2

schloerke (approved):
Thank you for the video!

PR #150, opened 2024-07-13
Avoid intercepting network requests in the service worker when we don't need to

georgestagg:
When we're not intercepting a Wasm app URL and not adding COI headers to a request, we can avoid a (potentially many MB) copy in the service worker by not intercepting the request at all.

This should reduce noise in the Network tab of web browsers' devtools, and provide a minor performance boost when fetching additional content for webR, such as `.../cairo.so`


wch (approved):
Good thinking!

PR #153, opened 2024-07-18
feat: Specific starting apps for the editor only view

gadenbuie:
The app template we use for https://shinylive.io/r/editor currently opens an app with a decent bit of code and comments (compare it with https://shinylive.io/py/editor).

As a heavy user of shinylive.io/r/editor, this adds a decent bit of friction because I need to delete most of the app before I can start working on my example.

This PR modifies the current example flow to add support for a special "Editor" category that, when present, is the starting code for the `{appEngine}/editor/` end point. In `{appEngine}/examples/` page, the Editor category is filtered out and won't appear in the sidebar.


georgestagg (approved):
This looks reasonable to me.

If we wanted to avoid the extra complexity we could also just add your minimal example as the first app, named something like `000-basic`. However, this way works OK too.

PR #157, opened 2024-07-19
Update shinyswatch to v0.7.0

### LINKED ISSUE #156: Update py-shinyswatch to v0.7.0
We released shinyswatch v0.7.0 yesterday before the shinylive v0.5.0 release, but shinylive includes shinyswatch v0.6.0. 

shinyswatch used to not be included in shinylive and we instructed users to add a `requirements.txt` with `shinyswatch` if they wanted to use shinyswatch in shinylive. However, that must have changed (looks like in #73) because now `requirements.txt` is ignored ([example shinylive app](https://github.com/posit-dev/shinylive/commit/78e015d71e7adf786492bb83f9c3a9eb12dd0d17)).

gadenbuie:
Fixes #156

Makes shinyswatch a submodule so that we can explicitly update the bundled version of shinyswatch.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -16,3 +16,6 @@
 [submodule "packages/plotnine"]
 	path = packages/plotnine
 	url = https://github.com/wch/plotnine.git
+[submodule "packages/py-shinyswatch"]
+	path = packages/py-shinyswatch
+	url = https://github.com/posit-dev/py-shinyswatch.git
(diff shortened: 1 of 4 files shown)


schloerke (approved):

wch:
One of the make targets will update the shinylive_lock.json file, so it’s not necessary to use a submodule here.

wch:
The make target is update_packages_lock. The reason it doesn’t automatically run with make all is so that the locked packages aren’t accidentally updated - I believe that updating some of them right now will cause shinylive to break. You can run that make target and then only keep the changes to the shinyswatch entry in the lock file.

wch (approved):

PR #158, opened 2024-07-19
draft: Add snippets

gadenbuie:
@georgestagg @wch In returning to #153, I wondered if the feature I'm actually wanting is _snippets_.

I took a stab at defining language-specific shiny snippets for R and Python, using the same snippets we use in the Shiny VS Code extension...

...and it _almost works_. I have the sense I'm fairly close ([drawing mainly from this example](https://discuss.codemirror.net/t/adding-custom-snippets/3535)), but the snippets aren't appearing in the editor. I'm hoping one of you might be able to quickly spot the issue.

### DIFF
--- a/src/Components/codeMirror/customSnippets.ts
+++ b/src/Components/codeMirror/customSnippets.ts
@@ -1,0 +1,126 @@
+import type { Completion } from "@codemirror/autocomplete";
+import { snippetCompletion } from "@codemirror/autocomplete";
+
+type ShinySnippet = { label: string; code: string[] };
+
+const shinySnippets: {
+  [key: string]: ShinySnippet[];
+} = {
+  r: [
+    {
+      label: "shinyapp",
+      code: [
+        "library(shiny)",
+        "library(bslib)",
+        "",
+        "ui <- page_${1:fluid}(",
+        "  ${0}",
+        ")",
+        "",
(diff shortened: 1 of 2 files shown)


georgestagg:
I've made a commit that gets things working for me. With this change, typing `shiny` in the editor offers the snippets as autocomplete options.

We could think about populating other properties like [`displayLabel`](https://codemirror.net/docs/ref/#autocomplete.Completion.displayLabel)⁠ alongside the `label`, so as to make the snippets clearer in the autocomplete list. Providing a `type` would give the entries an icon, for example. We could also think about setting `boost` to some number so that snippets float to the top.

src/Components/codeMirror/language-server/lsp-extension.ts
```diff
@@ -60,7 +60,6 @@ export function languageServerExtensions(
         }
       }
     }),
-    autocompletion(lspClient, filename),
```
georgestagg, line 63: I'm not sure why, but removing this definitely was required to get snippets working. I guess it's replacing the autocompletion extension setup by the file above.

In a quick test, removing this did not seem to do any harm, context aware completion still seemed to work. Still, we should try and work out why this was originally here and if it's still strictly necessary - though it doesn't seem to be.

PR #161, opened 2024-07-24
Use Pyodide's version of IPython

wch:
Recent versions of Pyodide come with IPython 8.23.0. Previous versions did not, and we needed to shim in a stub for IPython. This PR makes it use the Pyodide version of IPython. It also removes an entry in shinylive_lock.json that should not have been there (and didn't have an effect anyway, because our IPython shims took precedence over this entry.)

@cpsievert It might be worthwhile to run `make update_packages_lock` and see if anything breaks.

### DIFF
--- a/shinylive_lock.json
+++ b/shinylive_lock.json
@@ -343,25 +343,6 @@
       "tenacity"
     ]
   },
-  "ipython": {
-    "name": "ipython",
-    "version": "8.18.1",
-    "filename": "ipython-8.18.1-py3-none-any.whl",
-    "sha256": "e8267419d72d81955ec1177f8a29aaa90ac80ad647499201119e2f05e99aa397",
-    "url": "https://files.pythonhosted.org/packages/47/6b/d9fdcdef2eb6a23f391251fde8781c38d42acd82abe84d054cb74f7863b0/ipython-8.18.1-py3-none-any.whl",
-    "depends": [
-      {"name": "decorator", "specs": []},
-      {"name": "jedi", "specs": [[">=", "0.16"]]},
-      {"name": "matplotlib-inline", "specs": []},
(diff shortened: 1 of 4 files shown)


wch:
I've also updated all the packages in shinylive_lock.json.

shinylive_lock.json
```diff
@@ -481,133 +455,24 @@
...
-    "name": "idna",
-    "version": "3.6",
-    "filename": "idna-3.6-py3-none-any.whl",
-    "sha256": "c05567e9c24a6b9faaa835c4821bad0590fbb9d5779e7caa6e1cc4978e7eb24f",
-    "url": "https://files.pythonhosted.org/packages/c2/e7/a82b05cf63a603df6e68d59ae6a68bf5064484a0718ea5033660af4b54a9/idna-3.6-py3-none-any.whl",
-    "depends": [],
-    "imports": [
-      "idna"
-    ]
-  },
```
cpsievert, line 555: I'm not sure why, but when I run `make update_packages_lock` locally, it wants to add back in a lot of these dependencies. It also wants to move shiny from 1.0.0 to 0.10.2.9000 (even after I `cd packages/py-shiny && make install`). Maybe I'm missing something to run `make update_packages_lock` locally?

wch: First check if any other local package submodules are out of sync with the specified commit, with `git status`.

To update the version of shiny (and others packages) to the one pointed to by the submodule, run `make submodules`.

Then try running `make update_packages_lock`

cpsievert: Yea, the submodules are up-to-date, and I still get the differences described above.

cpsievert (approved):
Thank you! 🎉

PR #166, opened 2024-07-30
Post message to parent when ready

wch:
If there is a parent window, the `App` will now post a message to the parent when it is ready.

Notes:
- Currently, "ready" means that it is ready to receive `setFiles` messages posted to the this window, but conceivably one might want to have signals of other events (like when the app actually starts running), so it might make sense to have a message type that can represent multiple types of events.

### DIFF
--- a/src/Components/App.tsx
+++ b/src/Components/App.tsx
@@ -8,6 +8,7 @@ import {
 } from "../examples";
 import type { PyodideProxyHandle } from "../hooks/usePyodide";
 import { initPyodide, initShiny, usePyodide } from "../hooks/usePyodide";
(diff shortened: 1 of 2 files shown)


src/Components/App.tsx
```diff
@@ -329,6 +330,12 @@ export function App({
     };
   }, [currentFiles, setCurrentFiles, editorMethods]);
 
+  useRunOnceOnMount(() => {
+    if (window.parent === window) return;
+
+    window.parent.postMessage({ type: "shinyliveAppReady" }, "*");
```
jcheng5, line 336:
```suggestion
    window.parent.postMessage({ type: "shinyliveReady" }, "*");
```

jcheng5 (approved):

PR #187, opened 2024-11-08
feat: brand.yml support in shinylive for Python

gadenbuie:
This PR brings shinylive support for [brand.yml](https://posit-dev.github.io/brand-yml) with a number of updates:

1. First, we now bundle `libsass` in the pyodide distribution. I'm currently building libsass-python to pyodide in [GitHub Actions in my fork of libsass-python](https://github.com/gadenbuie/libsass-python/blob/dev/.github/workflows/pyodide.yml). It'd be nice to bring the package building into this repo. If that's easy for someone with more experience with pyodide tooling, I'd be happy for the help. Otherwise, I don't think it's worth the side-quest at the moment.

2. libsass is included in the pyodide distribution but requires `shiny[theme]` be declared in `requirements.txt`. This matches how theming works locally and reduces surprises. To make this work, I had to update our dependency installation to support installing package extras.

    Prior to this PR, including `shiny[theme]` in `requirements.txt` would only install `shiny`, not the extra `theme` requirements. Now, extra requirements are also installed. This isn't directly supported by `micropip`, so we manually look up the extra dependencies and install them directly.

    An informative message is printed when this happens, e.g.

    ```
    Installing libsass for shiny[theme]
    Installing brand-yml for shiny[theme]
    ```

3. I added a branded theming example to the examples list. Branded theming uses `.yml` and `.scss` files, so I also updated codemirror to include their language support.

4. Finally, I wanted to link to the brand-yml website from the example description, so I updated the Example React component to accept raw HTML in `about.txt`, which is sanitized via DOMPurify. We also add `target="_blank"` to any `<a>` elements.


shinylive_lock.json
```diff
@@ -97,7 +97,7 @@
   },
   "plotnine": {
     "name": "plotnine",
-    "version": "0.0.post20+g64295e1",
+    "version": "0.0.post1930+g64295e1",
```
wch, line 100: I think this change shouldn't be committed -- I believe your local copy of plotnine is getting `post1930` probably because it doesn't have the tags cloned, or something like that. (This is another case where the auto-numbering based on commits since the last tag is difficult to deal with.)

src/hooks/usePyodide.tsx
```diff
@@ -390,20 +391,63 @@ async def _install_requirements_from_dir(dir: str) -> None:
...
+                extra_reqs = [
+                    r for r in pkg_reqs
+                    if f'extra == "{extra}"' in str(r)
+                    or f"extra == '{extra}'" in str(r)
+                ]
+
+                for pkg_req in extra_reqs:
+                    pkg_req_name = re.sub(r"([a-zA-Z0-9._,-]+)(.*)", r"\\1", pkg_req).strip()
+                    if pkg_req_name not in micropip.list():
+                        print(f"Installing {pkg_req_name} for {pkg_name}[{extra}]")
```
wch, line 449: It would be slightly easier to understand if it was `extra_req_name` (and `extra_req` above).

examples/python/brand/Monda-OFL.txt
```diff
@@ -0,0 +1,93 @@
+Copyright 2021 The Monda Project Authors (https://github.com/googlefonts/mondaFont)
```
wch, line 1: (Commenting here because I can't comment on the .ttf file directly.)
Something to keep in mind for future examples: All of the examples right now are bundled into a single file, examples.json, and that whole file must be loaded before the page starts running an example. So that's a reason to try to example file sizes small. In the future, it might be better to put each example in its own .json file that is loaded on demand. I don't think it's a show stopper here, but it is something to keep in mind in the future when adding large examples.

shinylive_lock.json
gadenbuie: oh shoot, yeah I meant to do a `git add --patch` here but got carried away when adding the file to the commit, I'll fix it.

examples/python/brand/about.txt
```diff
@@ -0,0 +1,2 @@
+Branded Theming
+With a custom <a href="https://posit-dev.github.io/brand-yml">brand.yml</a>
```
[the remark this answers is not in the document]
gadenbuie: Yeah, I can get behind this. It'd be nice to have a better interface for additional content – sort of like the showcase mode – but I agree we should keep this simple for now.

examples/python/brand/Monda-OFL.txt
gadenbuie: I included this in part to test out the full range of features. Having tested it, I'm on the fence. I'd be okay taking it out for simplicity and sizing; or I'd equally be okay with leaving it so that we continue to test it manually before py-shiny releases. It adds about 200kb to `examples.json`, which is ~400kb before this change and just over 600kb after.

shinylive_lock.json
gadenbuie: Done in 2c4e1a26a20ca6d94522a5badb57dcbf7a731bba

src/hooks/usePyodide.tsx
```diff
@@ -390,20 +391,63 @@ async def _install_requirements_from_dir(dir: str) -> None:
...
+
+            pkg_reqs = dist.requires or []
+
+            for extra in valid_extras:
+                # Convert requires records  : 'libsass>=0.23.0; extra == "theme"'
+                # into just the package name: 'libsass'
+                extra_reqs = [
+                    r for r in pkg_reqs
+                    if f'extra == "{extra}"' in str(r)
+                    or f"extra == '{extra}'" in str(r)
```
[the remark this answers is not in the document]
gadenbuie: > Will they always have this exact formatting, with the spaces around `==`?

I think so, or at least if I've found the right documentation it appears that it will: https://packaging.python.org/en/latest/specifications/core-metadata/#provides-extra-multiple-use.

> It would be helpful if ... there was a comment illustrating what the item looks like

Maybe you were scanning quickly; there's exactly this comment two lines above, before `extra_reqs` is defined.

src/hooks/usePyodide.tsx
gadenbuie: Done in bbf9418e8bf796a60a865374b8995d4d8ef7b750

scripts/build_examples_json.ts
```diff
@@ -59,10 +59,20 @@ export default function buildExamples(examplesDir: string, buildDir: string) {
...
         })
         .sort((a: string, b: string) => {
-          // Sort files, with "app.py" first, and other files in normal sorted
-          // order.
-          if (a === "app.py" || a === "app.R" || a === "server.R") return -1;
-          if (b === "app.py" || b === "app.R" || b === "server.R") return 1;
+          // Sort files, with app related files first, followed by common
+          // support files, other files in normal sorted order.
+          const primary = ["app.py", "app.R", "server.R"];
+          const secondary = ["_brand.yml", "requirements.txt"];
```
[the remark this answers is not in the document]
gadenbuie: Sounds good, done in 1f3e43627d6b4aa81f7b2750aa619efd13327a7d

examples/python/brand/about.txt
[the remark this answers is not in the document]
gadenbuie: Done in cd71c010e6ed5702ba8037816d4b250b3963bbad and a523d69461e2a252dc5da779e37fcf923343dca6

gadenbuie:
@wch I've addressed feedback (thanks!) and linked to relevant commits in the comments above. The only thing I didn't fully resolve is the inclusion of the font file (https://github.com/posit-dev/shinylive/pull/187#discussion_r1835115136)

wch (approved):
Looks good, thanks for addressing the comments! I think the font thing is OK for now -- we'll just want to be mindful of the examples.json size in the future.

wch:
@gadenbuie after it gets built and deployed to github.io, feel free to move the `deploy` branch to point to the same commit as `main`.

PR #206, opened 2025-04-08
feat: Disable server-side bookmarking for py-shiny

schloerke:
Related: https://github.com/posit-dev/py-shiny/issues/1914

Related: https://github.com/posit-dev/py-shiny/issues/1956

--------------

Q: Should [R shiny](https://github.com/posit-dev/shinylive/blob/0ca09681baf8116307e952935c62a3a5ec44452b/src/hooks/useWebR.tsx#L149) also have it's server-side bookmarking disabled?
  * Can R perform server-side bookmarking?

Q: This might get the app to have bookmarking... but what about forwarding any query string values to the parent frame for shinylive.io? If the same answer as r-shiny is... not at the top of the priority queue... then we can wait until later!

TODO:
- [x] Test locally (cc @karangattu )

### DIFF
--- a/src/hooks/usePyodide.tsx
+++ b/src/hooks/usePyodide.tsx
@@ -377,8 +377,23 @@ async def _start_app(app_name, scope = _shiny_app_registry, dev_mode = False):
     import os
     import sys
     import importlib
+    import shiny
     import shiny.express
     from pathlib import Path
+
+    if (
+        hasattr(shiny, "bookmark")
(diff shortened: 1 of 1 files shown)


karangattu:
Verified that enabling server side bookmarking for an app in shinylive throws a error notification - `Error bookmarking state: shinylive is not configured to save sessions to disk.`

When using url bookmarking, the query params still do not update in the url address bar.

PR #211, opened 2025-07-28
Update jest

wch:


jat255:
I think this can be closed now that #246 was merged

PR #234, opened 2026-07-29
ci: Check that every example is listed in examples/index.json

schloerke:
Adds a CI check that `examples/index.json` and the `examples/` directories agree, then removes the example directories that check turns up.

Split into two commits on purpose: the first adds the check and **leaves CI red**, so the failure is visible in the run history; the second deletes the orphans and takes it green.

## Why

`examples/index.json` drives which apps get built into `examples.json`, and that file is the only example content that ships — the raw `examples/` tree is never copied into `build/`. So an app directory nobody lists is:

- never built into `examples.json` (`build_examples_json.ts` maps over the ordering, it never reads the directory),
- never served, and
- **not loadable by URL** — `allowExampleUrl` resolves names via `examples.json` (`App.tsx`), and on a miss it silently falls back to the first example rather than erroring.

That silent fallback is why nobody noticed. 23 Python example directories were unreachable, most of them since #94 (2024-01-29), which pruned whole categories out of `index.json` but left every directory on disk.

The cost wasn't just dead weight: these files kept getting *maintained*. Later commits swept through them removing `@output` and switching to `render.code` — real effort spent on apps no user could load. Two of them had also rotted outright, using `ui.nav` / `ui.panel_sidebar` / `ui.panel_main`, all removed from Shiny.

## What the check covers

Both directions, plus a third case:

| Case | Result |
| --- | --- |
| Extra on disk, not in `index.json` | ✖ lists each orphan |
| In `index.json`, missing on disk | ✖ lists each missing app |
| Indexed and present, but no `about.txt` | ✖ lists each one |

`build_examples_json.ts` already throws for a missing indexed app, but only part way through `make all` and only on the first one; this reports every offender up front. The `about.txt` case was genuinely unguarded — `parseApp()` reads it unchecked, so a new example without one fails the build with a bare ENOENT.

Engines come from `index.json` rather than being hardcoded, so both `python` and `r` are covered, and it errors if `index.json` names an engine directory that doesn't exist.

## Notes

- The script is **dependency-free** — node builtins only, no TypeScript — so the job runs straight from a checkout with no `npm ci`. Verified by running it against a bare `git archive` export with no `node_modules` present.
- It's in a new `test-apps.yml` rather than `build.yml`, and has no `needs` relationship to `build`. A broken example reports its own red check but cannot block a build or a deploy. More example-level tests can go here.
- Also available locally as `make check_examples_index`.


schloerke:
Merging as I can't see any gain in keeping the unused apps. Easy to revert later if we want them back.

cc @cpsievert

PR #236, opened 2026-07-29
test: Smoke test every example app, and fix what it found

### LINKED ISSUE #231: test: Add smoke tests for all examples
Similar to the py-shiny website.

I'm guessing we'll just add a new make target(s) to install playwright and run it through a similar script.

schloerke:
Closes #231.

Loads all 39 indexed examples (28 Python, 11 R) in headless chromium and fails on tracebacks, deprecation warnings, Shiny output errors, or JS console errors. Then fixes everything it found. **CI is green.**

## What the test caught

Three real, pre-existing bugs — each failed deterministically on both the initial attempt and the retry, before being fixed in `5455bb4`:

| Example | Failure |
|---|---|
| `plot_interact_basic` | `app.py:30 ShinyDeprecationWarning: panel_well() is deprecated` |
| `plot_interact_select` | `app.py:27 ShinyDeprecationWarning: panel_well() is deprecated` |
| `examples/r/008-html` | `404 /r/examples/app_.../shared/shiny.css` — the file does not exist in the webR image, so the app rendered unstyled. Upstream renamed these to `.min.` in rstudio/shiny#4220. |

## Also cleaned up (no warning, but stale)

- `ui.row()`/`ui.column()` → `ui.layout_columns(col_widths=)` in the three `plot_interact` examples and `static_content`. These were the last uses in the directory. `plot_interact_exclude`'s `ui.column(2)` spacer became `col_widths=[-2, 8, -2]`, since negative widths are empty columns.
- Dropped the redundant bare `@output` in `static_content` (unnecessary since shiny 0.6). The one left in `cpuinfo` passes `suspend_when_hidden=False`, so it stays.
- Replaced `file_download_core`'s hand-rolled `ui_card()` div helper with `ui.card()`/`ui.card_header()`.
- Deleted `plot_interact_select`'s `format_table()`, dead since it was defined after `app = App(...)`.

`@render.download` is deliberately unchanged: `@render.download_button` looks like the modern replacement but **does not exist in shiny 1.6.4**, which is what `shinylive_lock.json` pins and therefore what these apps actually run against.

## Design notes

**Why not reuse `playwright.config.ts`** — its `webServer` entries shell out to `shinylive export`, which needs the Python `shinylive` package, which depends on this repo. That circular dependency is why the playwright job in `build.yml` has been commented out for years. This suite serves the built `_shinylive/` with `python3 -m http.server` instead: no Python package, and it exercises the bytes we deploy. `playwright.config.ts` gets a `testIgnore` so `make test` is unaffected.

**Why a fresh page per example** — walking one warm session races the outgoing app: in-flight proxied requests hit a `_shiny_app_registry` key that has already been swapped (Pyodide) or take a null channel response (webR). Loading those same apps directly is clean, so both are artifacts of switching. A cold boot off a local file server is cheap, so isolation also removes any chance of one broken app cascading into the rest — which had masked 3 R examples behind the `008-html` failure.

**Terminal triage is a deny-list** of failure signatures, not an allow-list of benign output, because normal chatter is open-ended. `SEND:`/`RECV:` frames are allowed through: several examples set `App(debug=True)` and those frames embed an `"errors": {}` key plus arbitrary app values. This check is weaker for R than Python — Pyodide wires `stderr` to the terminal, webR writes nothing — so R leans on the output-error and console-error assertions.

## CI cost

Started at ~12 min per job, run **twice** per push (an unfiltered `push` alongside `pull_request`, same commit). Measured breakdown was 91s build / 544s tests, so caching the build would have recovered almost nothing.

Now: `push` scoped to `main`, a `concurrency` group cancelling superseded runs, and the suite split by engine then sharded 3 ways each — 6 jobs, one runner apiece, since these want memory rather than cores. Measured on `99a02b4`:

```
examples-check-index         5s
examples-smoke-test (py, 1) 247s
examples-smoke-test (py, 2) 217s
examples-smoke-test (py, 3) 210s
examples-smoke-test (r, 1)  198s
examples-smoke-test (r, 2)  194s
examples-smoke-test (r, 3)  176s
```

| | wall | runs/push | total runner-secs |
|---|---|---|---|
| original | 12.0 min | 2 | ~1270 |
| 4-way shard | 5.8 min | 1 | ~1080 |
| engine + shard | **4.1 min** | 1 | ~1250 |

**2.9x faster wall-clock, at slightly less total runner time** than the original, since dropping the duplicate trigger paid for the extra jobs.

Splitting by engine before sharding matters: R examples cost ~29s each on CI against ~17s for Python, and playwright assigns shards in contiguous blocks, so sharding the combined suite piled all 11 R tests into one shard that ran 45% longer than three idle siblings.

Two things worth knowing if you touch this:

- **Sharding is file-granular unless `fullyParallel: true`.** This suite is one file, so shard 1 would take every test and the rest none — green CI with most of the work silently skipped.
- **A project's `grep` target includes the project name and the spec's path.** Under project `r`, `grep: /r examples/` also matched every Python test via the `r examples-smoke.spec.ts` in its target. Describe titles carry an `engine:` marker instead.

Verified the six jobs are an exact cover: union of their test lists is 38 titles, no duplicates, nothing missing.

## Make targets

- `make examples-check-index` (renamed from `check_examples_index`)
- `make examples-smoke-test` (new; `EXAMPLES_SHARD=1/4` runs one shard)

Lives in `test-apps.yml` with no `needs` on the build, so a broken example still cannot block a build or a deploy.

Filed #237 separately for an unrelated thing this turned up: a synchronous `metadata.rds` probe that re-runs on every R app start and 404s.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,10 +23,10 @@ jobs:
       fail-fast: false
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v7
 
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@v7
         with:
           python-version: ${{ matrix.python-version }}
 
@@ -54,7 +54,7 @@ jobs:
       # from making a release. To fix this, we will probably need to avoid
       # installing the Python package.
 
-      # - uses: actions/setup-node@v3
+      # - uses: actions/setup-node@v7
       #   with:
       #     node-version: "14.x"
       # - name: Install dependencies
@@ -65,7 +65,7 @@ jobs:
       #   run: npx playwright install --with-deps
       # - name: Run Playwright tests
       #   run: make test
-      # - uses: actions/upload-artifact@v3
+      # - uses: actions/upload-artifact@v7
       #   if: always()
       #   with:
       #     name: playwright-report
@@ -104,7 +104,7 @@ jobs:
       # =====================================================
       - name: Upload _shinylive/ artifact
         if: github.ref == 'refs/heads/main'
-        uses: actions/upload-pages-artifact@v3
+        uses: actions/upload-pages-artifact@v5
         with:
           path: "_shinylive/"
 
@@ -155,4 +155,4 @@ jobs:
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@v4
+        uses: actions/deploy-pages@v5
--- a/.github/workflows/test-apps.yml
+++ b/.github/workflows/test-apps.yml
@@ -8,11 +8,50 @@ on:
   pull_request:
 
 jobs:
-  check_examples_index:
+  examples-check-index:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v5
 
       - name: Check that all examples are indexed
         run: |
-          make check_examples_index
+          make examples-check-index
+
+  # Loads every indexed example in a real browser, for both engines, and fails
+  # on tracebacks, deprecation warnings, output errors, or JS console errors.
+  #
+  # This has to build shinylive first, because it drives the built `_shinylive/`
+  # output. It deliberately does not reuse the artifact from build.yml: this
+  # workflow has no `needs` relationship to that one, so a broken example still
+  # cannot block a build or a deploy.
+  examples-smoke-test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.11"
+
+      - name: Upgrade pip
+        run: python -m pip install --upgrade pip
+
+      - name: Check out submodules
+        run: |
+          make submodules
+
+      - name: Build shinylive
+        run: |
+          make all
+
+      - name: Run examples smoke test
+        run: |
+          make examples-smoke-test
+
+      - uses: actions/upload-artifact@v4
+        if: always()
+        with:
+          name: examples-smoke-test-report
+          path: playwright-report/
+          retention-days: 30
--- a/.github/workflows/test-apps.yml
+++ b/.github/workflows/test-apps.yml
@@ -4,9 +4,6 @@ name: Test example apps
 # so they live in their own workflow: a problem with an example should not block
 # a build or a deploy.
 on:
-  # `pull_request` alone covers branch work. Listing an unfiltered `push` as well
-  # would run this whole workflow twice for every push to a PR branch, on the
-  # same commit.
   push:
     branches: [main]
   pull_request:
@@ -34,17 +31,21 @@ jobs:
   # workflow has no `needs` relationship to that one, so a broken example still
   # cannot block a build or a deploy.
   #
-  # The suite is sharded so each shard has a runner to itself -- every test boots
-  # a whole Pyodide or webR instance, so they want memory rather than cores, and
-  # sharing one runner between parallel workers would squeeze all of them. Four
-  # shards puts the slowest job around four minutes; past that the per-shard
-  # build cost stops paying for itself.
+  # Split by engine and then sharded, so every job has a runner to itself --
+  # each test boots a whole Pyodide or webR instance, so they want memory rather
+  # than cores, and parallel workers inside one runner would squeeze all of them.
+  #
+  # Splitting by engine first matters because R examples cost roughly 29s each
+  # against 17s for Python, and playwright assigns shards in contiguous blocks.
+  # Sharding the combined suite therefore piled all 11 R tests into the last
+  # shard, which then ran 45% longer than its siblings while they sat idle.
   examples-smoke-test:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        shard: [1, 2, 3, 4]
+        engine: [py, r]
+        shard: [1, 2, 3]
     steps:
       - uses: actions/checkout@v5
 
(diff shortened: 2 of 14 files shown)


schloerke:
## CI result: works as intended

[Run 30415891331](https://github.com/posit-dev/shinylive/actions/runs/30415891331) — `examples-check-index` ✅, `examples-smoke-test` ❌ with **35 passed, 3 failed**, identical to local. Each of the three failed on both the initial attempt and the retry, so these are deterministic, not flake.

```
Basic plot interaction: unexpected output in the shinylive terminal
  + "app.py:30: ShinyDeprecationWarning: panel_well() is deprecated. Use shiny.ui.card() instead."
Selecting data: unexpected output in the shinylive terminal
  + "app.py:27: ShinyDeprecationWarning: panel_well() is deprecated. Use shiny.ui.card() instead."
Custom HTML UI: browser console errors
  + "Failed to load resource: the server responded with a status of 404 ()"
```

Traced the R 404 to the app, not the infrastructure:

```
404 /r/examples/app_4ako77mqsjh399skm3px/shared/shiny.css
```

## Timing

Smoke test 9.9 min, build ~2.5 min, job total ~12.5 min. Slower than the ~5 min local run, mostly engine boots on a 2-core runner. If that's too slow for every push, the obvious lever is caching `downloads/` so `make all` skips re-fetching Pyodide and webR.

## Unrelated observation

Every R example triggers `HEAD /r/shinylive/webr/packages/metadata.rds → 404`. webR handles it without raising a console error, so nothing fails, but it is one wasted round trip per R app start and may be worth a separate look.

schloerke:
## CI cost, final

Split by engine then sharded 3 ways each. Measured on `99a02b4` ([run 30419312643](https://github.com/posit-dev/shinylive/actions/runs/30419312643)), all green:

```
examples-check-index         5s
examples-smoke-test (py, 1) 247s
examples-smoke-test (py, 2) 217s
examples-smoke-test (py, 3) 210s
examples-smoke-test (r, 1)  198s
examples-smoke-test (r, 2)  194s
examples-smoke-test (r, 3)  176s
```

| | wall | runs/push | total runner-secs |
|---|---|---|---|
| original | 12.0 min | 2 | ~1270 |
| 4-way shard | 5.8 min | 1 | ~1080 |
| engine + shard | **4.1 min** | 1 | ~1250 |

**2.9× faster wall-clock than where this started, at slightly less total runner time**, since dropping the duplicate `push` trigger paid for the extra jobs.

Spread is now 176–247s with Python as the long pole, versus the previous 235–348s where one R shard ran 45% longer than three idle siblings.

## Two non-obvious things, in case they bite someone later

**A project's `grep` is matched against a target that includes the project name and the spec's path**, not just the describe and test titles. Under project `r`, a `grep` of `r examples` matched every *Python* test as well — via the `r examples-smoke.spec.ts` in its target — selecting all 38 instead of 11. Describe titles now carry an `engine:` marker that cannot collide.

**Sharding is file-granular unless `fullyParallel: true`.** This suite is a single file, so shard 1 would have taken every test and the rest none — green CI with most of the work silently skipped. Checked with `--list` rather than assumed.

Verified the six jobs are an exact cover: the union of their test lists is 38 titles, no duplicates, nothing missing.

Per review, the comment above the `on:` triggers is gone, and the `paths:` filter is skipped for now.

schloerke:
cc @cpsievert Apps have been modernized

schloerke:
## Node.js 20 deprecation cleared

`upload-artifact@v4` was the one being reported, but not the only offender — `actions/deploy-pages@v4` in `build.yml` is also `runs.using: node20`. It just never surfaced, because `deploy_gh_pages` is skipped on pull requests. Read `runs.using` from each action at its pinned ref rather than guessing:

```
actions/checkout              v5 -> node24   (fine, but not latest)
actions/setup-python          v6 -> node24   (fine, but not latest)
actions/upload-artifact       v4 -> node20   ** reported
actions/deploy-pages          v4 -> node20   ** unreported
actions/upload-pages-artifact v3 -> composite
```

All `actions/*` now on their latest major: checkout v7, setup-python v7, upload-artifact v7, upload-pages-artifact v5, deploy-pages v5 — plus the two commented-out actions in build.yml's disabled playwright block, so they don't get resurrected several majors behind.

### Verified rather than assumed

Before/after annotations on the smoke jobs:

```
30419312643 (v4): [warning] Node.js 20 is deprecated. ... actions/upload-artifact@v4
30419774922 (v7): (no warnings on any job)
```

`build.yml` on the same commit: **no warnings either**, so `deploy-pages@v5` and `upload-pages-artifact@v5` are clean too.

Both workflows green. Smoke timings unchanged: py 266/242/186s, r 209/206/170s.

### Behavioural changes I checked, since the deploy path can't run from a PR

- `upload-artifact` v5's "BREAKING CHANGE" is only the Node bump; v6 requires runner ≥ 2.327.1, which GitHub-hosted runners satisfy.
- `upload-pages-artifact` v4 **stopped including hidden files**. The only dotfiles under `_shinylive/` are six `.gitignore` files copied in from `site/`, none of which the deployed site needs. `include-hidden-files` exists if that ever changes.
- `checkout` v7 blocks checking out a fork PR under `pull_request_target` and `workflow_run`. `build.yml` *is* triggered by `workflow_run`, but its checkout takes no `ref`, so it is unaffected.
- Confirmed every input/output still in use exists in the new versions: `upload-artifact`'s `name`/`path`/`retention-days`, `upload-pages-artifact`'s `path`, `deploy-pages`' `page_url` output, `setup-python`'s `python-version`.

### Left alone — two third-party actions are still node20

Outside `actions/*`, so not touched, but they will keep warning once the reporting widens:

| action | latest | note |
|---|---|---|
| `aws-actions/configure-aws-credentials@v5` | v6 is node24 | sits on the S3 deploy path; untestable from a PR |
| `actions4git/setup-git@v1` | v1.0.3 is newest | no node24 major exists yet |

Happy to bump the AWS one in a separate PR where it can be watched on a real deploy.

Issue #192, opened 2024-12-05
Long loading times with the shinylive examples - maybe room for tree-shaking or lightweight r engines?


joanbadia-IISPV:
Hello,
First of all, I love what you're doing! 

However, I noticed that in general, the loading times of the web app are a bit long, and I wonder if it's because the "bloatware" associated with R or Python.

Are you guys trying to implement any tree-shaking method, like the ones commonly used in web-apps, to remove all non-used code and leave a lightweight engine? I am aware that for that you need a dependency tree, but I think recently, the positron team achieved something similar? Not 100% sure.

Or alternatively, do you have a lightweight R engine implementation?

Thank you again!

wch:
As far as I know, tree shaking Python or R code isn't really feasible -- both languages do so much dynamic stuff that it's impossible to reliably detect whether any particular piece of code will be reached.

PR #240, opened 2026-07-29
feat: Add explanatory loader text is during first load and handle potential loading failure modes

jat255:
## Problem

On a first visit to a site using shinylive, the app-loading overlay shows the pulsing hex animation with no text. Behind the scenes, a multi-megabyte Pyodide or webR engine is downloaded plus package installs, which can take tens of seconds (or longer on slow connections), so a slow first load doesn't give any feedback about if something useful is happening during that time. Generally speaking, I wanted to address "could the shinylive loader show a bit more help text on first load?"

Related: #192 (long loading times) describes the underlying slowness, which this does not attempt to fix.

## What changed

Two things, primarily:

**1. The overlay says what it is doing.** Under the existing hex animation:

- `Downloading Python…` → `Starting Python…` → `Loading packages and starting app…`
- and the R equivalents (`Downloading R…`, `Starting R…`)

The text only appears after a **3 second delay**, so a fast load (from cache) looks essentially as it did before.

A store implemented per engine in `src/load-status.ts` keeps track of the various loading stages.

**2. Failures produce an error screen instead of just endlessly pulsing.** 

The following conditions were tested:

| failure | before | after |
| --- | --- | --- |
| Engine wasm missing, or host unreachable | spun forever, no error | error in ~0.2s, naming the URL and the status |
| Worker reports an init failure (e.g. `pyodide-lock.json` 404) | discarded; surfaced later as an unrelated error | the real cause, in under a second |
| R app fails to start (e.g. syntax error) | app marked running, blank frame, error only in the terminal — and **nothing at all in an embed** | `Error starting app!`, naming the line and showing the offending source line |
| any of the above | traceback only | leads with hard-refresh / clear-cache advice |

Specifically, these limitations came from:

- **Engine wasm.** Neither `loadPyodide()` nor webR's `init()` settles when the core wasm app cannot be instantiated; its promise is neither resolved nor rejected, so there were no error signals the page can catch. `src/engine-load-guard.ts` checks that at least the wasm file is fetchable before the engine starts, and fails early if not.
- **`WebWorkerPyodideProxy.init()`** discarded the worker's error reply, so a Pyodide that never started would appear successful.
- **R app startup.** `.start_app` (in `useWebR.tsx`) is evaluated with `captureConditions = FALSE`: with this setting an R error makes `evalR` *resolve* rather than reject, so a raised error would go to the terminal and never reach JavaScript. `.start_app` therefore reports its outcome by returning a value. It returns `list(status = "ok")`, or `list(status = "error", message, class, call)` on failure. An earlier revision of this PR returned `""` for success and the message for failure, which made a condition carrying an empty message indistinguishable from success; the status list removes that. `Viewer.tsx` reads it with `evalR` + `toJs()` through the shelter it already opens and purges, rather than `evalRString`. The condition's `class` now reaches the console for diagnosis, and `call` is reported only when it names the author's own code — shiny wraps every app body in `..stacktraceon..()`, so a top-level failure's raw call is the whole app source, and that wrapper is filtered out. Flipping `captureConditions` to `TRUE` would also have fixed the original problem, but it changes error handling for every R call rather than just app startup. Package installation stays non-fatal on purpose — see the R note under Testing.
- **R syntax errors get a line number.** shiny's `sourceUTF8` catches the parse error and raises only `Error sourcing <file>`, discarding the detail. `.start_app` now quickly parses the app's top-level `.R` files before shiny sees them, so the real parse error including line, column, offending line and caret, reaches the viewer. It runs before the VFS mount and package installs, so a typo fails in a couple of seconds instead of after work the app cannot use.
- **User-facing advice.** There is deliberately no reload button: a reload triggered from JS cannot bypass the cache, so the error instead suggests the user to hard-refresh, as that is a possible solution in some cases.

### One known failure-mode limitation

An engine asset that downloads but cannot be *instantiated*, for example a truncated file, or an HTML error page served with 200, will still cause the loading hexes to spin with no error screen. Closing it needs either byte-level validation or a timeout. I had a timeout implemented, but removed it because it was a messy catch for a rare problem and definitely overengineered

## Architecture

The `load-status.ts` observable store controls the "status" of the engine and app load. `App.tsx` writes into the store via `initPyodide` or `initWebR`, and then `Viewer.tsx` responds to changes in the store to control what's shown to the user:

```mermaid
flowchart TB
    A["<b>App.tsx</b><br/>starts engine init, once per page"]
    B["<b>initPyodide / initWebR</b><br/>guard the asset, then boot the engine"]
    S["<b>load-status.ts</b><br/>one store per engine"]
    H["<b>useLoadStatus.ts</b><br/>useSyncExternalStore"]
    V["<b>Viewer.tsx</b>"]
    L["LoadingStatus.tsx<br/>stage text, after 3s"]
    E["ViewerError<br/>failure screen"]

    A --> B
    B -- "set(stage)" --> S
    S -- "subscribe" --> H
    H --> V
    V -- "loading" --> L
    V -- "failed" --> E

    classDef store fill:#fde8c8,stroke:#b8860b
    class S store
```

Stages are linear and one-way:
`engine-download` → `engine-start` → `ready`, or `failed`, which is terminal and causes an error to be displayed in the UI.

### How a failure reaches the screen

Failures were previously just silently discarded and the loading icon would endlessly pulse. Now, error sources fan into two distinct types of errors:

```mermaid
flowchart LR
    G["<b>guard</b><br/>wasm 404, host unreachable,<br/>not a wasm module"]
    W["<b>worker</b><br/>pyodide init error reply<br/><i>pyodide-proxy.ts:388</i>"]
    P["<b>engine init</b><br/>load_python_pre / load_r_pre,<br/>initShiny / initRShiny throws"]
    PA["<b>Python app</b><br/>_start_app throws<br/><i>Viewer.tsx:381</i>"]
    RA["<b>R app</b><br/>.start_app returns a status list<br/><i>Viewer.tsx:331</i>"]
    C["<b>cosmetic</b><br/>terminal clear, start banner"]

    F["stage = <b>failed</b><br/><i>terminal</i>"]
    X["appRunningState<br/>= <b>errored</b>"]
    LOG(["console.error only<br/><i>app keeps running</i>"])

    E1["<b>ViewerError kind='engine'</b><br/><i>+ recovery hint</i>"]
    E2["<b>ViewerError kind='app'</b><br/><i>traceback only</i>"]

    G --> F
    W --> F
    P --> F
    PA --> X
    RA --> X
    C --> LOG
    F --> E1
    X --> E2

    classDef prior fill:#f2f2f2,stroke:#aaa,color:#666
    classDef dead fill:#eef6ee,stroke:#8ab88a,color:#456
    class PA prior
    class LOG dead
```

The python app failure was the only one that already reached the screen before this PR. Everything else was silent.

The engine/app split determines the advice that gets shown to the user. The hard-refresh / clear-cache hint appears *only* on the engine side. On the app side the engine already loaded fine, so the cache isn't the suspect and the hint would point the reader away from the traceback sitting right below it (commit `6f99001`).

## Example Recordings

The following screen recordings demonstrate what this looks like in a few different scenarios, both with throttling to simulate a slow connection, and on a fast connection. There are both R and Python examples here, in an "embed"/Quarto type display, as well as the full screen height version like on https://shinylive.io/py/examples/ 

<table>
<thead>
<tr><th></th><th>Python</th><th>R</th></tr>
</thead>
<tbody>
<tr>
<td><strong>Full page &mdash; throttled</strong></td>
<td>
<video src="https://github.com/user-attachments/assets/df799a68-8d87-4c11-a3dc-bbd6af40e53c" controls width="440"></video>
</td>
<td>
<video src="https://github.com/user-attachments/assets/ca1bdc24-e240-4ce9-856c-0035e3a50890" controls width="440"></video>
</td>
</tr>
<tr>
<td><strong>Full page &mdash; unthrottled</strong></td>
<td>
<video src="https://github.com/user-attachments/assets/2c99b6ff-14fc-4984-bc32-3d47fc50144f" controls width="440"></video>
</td>
<td>
<video src="https://github.com/user-attachments/assets/7d3df46c-2613-44a1-8f1d-c974192755c5" controls width="440"></video>
</td>
</tr>
<tr>
<td><strong>Embedded (Quarto) &mdash; throttled</strong></td>
<td>
<video src="https://github.com/user-attachments/assets/85c3f9bb-a503-4778-8e83-63edaf94d625" controls width="440"></video>
</td>
<td>
<video src="https://github.com/user-attachments/assets/3326d7fd-dd82-40b1-8188-df4c34786530
" controls width="440"></video>
</td>
</tr>
<tr>
<td><strong>Embedded (Quarto) &mdash; unthrottled</strong></td>
<td>
<video src="https://github.com/user-attachments/assets/a4b2c7d3-0a25-42e7-96fa-631351384457" controls width="440"></video>
</td>
<td>
<video src="https://github.com/user-attachments/assets/9f74c1e3-2f8d-419a-8b85-5172d31752fc" controls width="440"></video>
</td>
</tr>
</tbody>
</table>

## Testing

Both test layers this PR needed landed while it was open: #246 added the jest unit harness and a `tsc --noEmit` CI gate, and #247 moved browser testing to pytest + Playwright under `tests/`. This PR uses both, so the manual verification below is now automated.

**Unit tests — 37, jest.** `src/load-status.test.ts` covers the status store: subscription semantics that `useSyncExternalStore` depends on, per-engine isolation, and `failed` being terminal. `src/engine-load-guard.test.ts` covers the reachability and wasm-magic-number checks against a stubbed `fetch`, including a magic number split across chunks, a body that cannot be cancelled, and one that throws mid-read. Coverage over the modules these import is 100% on all four metrics.

**Browser tests — 19, pytest + Playwright, `-m loader`.** `tests/test_loader_status.py` drives every mode in the table below end to end against the built site, for both engines, in both the full-page and embedded layouts. App source travels in the URL hash, so app-level failures need no interception; engine-level ones use `page.route()`. Two choices worth a reviewer's eye: the staged-loading test records transitions with an in-page `MutationObserver` rather than polling assertions, because a delay injected in a sync route handler blocks Playwright's own connection thread; and the delay is a `--loader-delay` option rather than CDP throttling, so assertions are on the observed order of stages and never on elapsed time.

`scripts/loader-demo.ts`, the local script behind the recordings above, is deleted. `tests/README.md` documents the equivalent recipes — `--headed --slowmo`, `--video on`, `--tracing on` with `playwright show-trace`, and turning `--loader-delay` up to watch the stages by hand.

The table below was the original manual verification; every row now has a test.

| mode | Python | R |
| --- | --- | --- |
| healthy load | app runs image| app runs image |
| core wasm 404 | error in 0.2s, names URL + status image | same image |
| connection dropped for the core wasm | error in 0.2s, reported as unreachable image | same image |
| syntax error in app source | `Error starting app!` + traceback image | `Error starting app!` + `app.R:7:1: unexpected symbol` with the offending line and caret **(new — previously silent in an embed)** image |
| unresolvable dependency | `Error starting app!` + micropip's `ValueError` image | warnings only; the app genuinely runs — see below image |

**On the R dependency case.** R has no `requirements.txt`; dependencies are inferred from the source by `renv::dependencies()`, and a missing one is tolerated. `library()` is shimmed by webR (`environmentName` reports `webr`) and returns normally instead of erroring, so the app really does start. That appears to be upstream's intended behaviour, which may also be why `.start_app` does not treat a failed install as fatal — `renv::dependencies()` reports packages that are never used (a `library()` call inside `if (FALSE)` is reported, and that app runs fine), so failing on it could break working apps.

**CI placement.** The loader tests carry both the `site` and `loader` markers, and the Make targets split on that: `make test-site` is `-m "site and not loader"`, `make test-loader` is `-m loader`. Between them they cover every `site` test, so running one is not running them all.

`make test-site` stays where it was, in `build.yml`'s build job, so a failure in the editor, URL-loading or static-export tests still stops the deploy that follows. The loader tests move to a new `test-loader.yml`, for the same reason `test-apps.yml` exists: they are slow — nearly every one boots a real Pyodide or webR instance, so the suite costs about as much as all the other site tests together — and they should not sit in front of an S3 deploy. Being a separate workflow also gets them a `cancel-in-progress` concurrency group, which `build.yml` cannot have without risking a cancelled deploy.

They therefore do not gate anything yet, which is deliberate and worth revisiting once they have some green history. CI has already earned that caution: the first run of this branch failed on an intermittent console-error assertion that passes locally and passed on the next run, which is exactly what gating would have turned into a blocked deploy. That bug is fixed — the test was collecting console errors itself after opting out of `fail_on_page_errors`, but had reimplemented the collection without the fixture's forgiveness for the tag-manager requests `block_analytics` aborts, which only exist where `GOOGLE_TAG_MANAGER_ID` is set. The predicate now lives in one place and both callers share it.

Moving the gate later means either folding the marker back into `make test-site`, or triggering the deploy from `test-loader.yml`'s success via `workflow_run`.

**Verification.** `make type-check` clean; `npm run test:unit` 328/328; `npm run test:unit:coverage` 100% on all four metrics; `venv/bin/pytest tests -m site` 34/34.

### DIFF
--- a/src/hooks/usePyodide.tsx
+++ b/src/hooks/usePyodide.tsx
@@ -50,7 +50,6 @@ export async function initPyodide({
   if (unreachable) throw new Error(unreachable);
 
   const pyodideProxy = await loadPyodideProxy(
-    // indexURL is Pyodide's own loadPyodide() parameter name; keep it as-is.
     { type: proxyType, indexURL: baseUrl },
     stdout,
     stderr,
@@ -144,8 +143,8 @@ export function usePyodide({
       const pyodideProxyHandle = await pyodideProxyHandlePromise;
       setPyodideProxyHandle(pyodideProxyHandle);
     })().catch((e) => {
-      // Already surfaced to the user via the load status store; log it so it
-      // isn't an unhandled rejection.
+      // Already surfaced to the user via the load status store;
+      // logged to console so it isn't an unhandled rejection.
       console.error(e);
     });
   }, [pyodideProxyHandlePromise]);
--- a/src/hooks/usePyodide.tsx
+++ b/src/hooks/usePyodide.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect } from "react";
+import { loadStatusStore } from "../load-status";
 import type { ProxyType, PyodideProxy } from "../pyodide-proxy";
 import { loadPyodideProxy } from "../pyodide-proxy";
 import * as utils from "../utils";
@@ -36,6 +37,9 @@ export async function initPyodide({
   if (!stdout) stdout = async (x: string) => console.log("pyodide echo:" + x);
   if (!stderr) stderr = (x: string) => console.error("pyodide error:" + x);
 
+  const status = loadStatusStore("python");
+
+  status.set("engine-download");
   const pyodideProxy = await loadPyodideProxy(
     {
       type: proxyType,
@@ -47,10 +51,15 @@ export async function initPyodide({
 
   let initError = false;
   try {
-    // One-time initialization of Python session
+    // One-time initialization of Python session. This await also loads the base
+    // packages, since the worker calls loadPackagesFromImports before running
+    // the code, so boot and package loading are reported as a single stage.
+    status.set("engine-start");
     await pyodideProxy.runPyAsync(load_python_pre);
+    status.set("ready");
   } catch (e) {
     initError = true;
+    status.set("failed", e instanceof Error ? e.message : String(e));
     console.error(e);
   }
 
@@ -127,7 +136,11 @@ export function usePyodide({
     (async () => {
       const pyodideProxyHandle = await pyodideProxyHandlePromise;
       setPyodideProxyHandle(pyodideProxyHandle);
-    })();
+    })().catch((e) => {
+      // Already surfaced to the user via the load status store; log it so it
+      // isn't an unhandled rejection.
+      console.error(e);
+    });
   }, [pyodideProxyHandlePromise]);
 
   return pyodideProxyHandle;
--- a/src/hooks/usePyodide.tsx
+++ b/src/hooks/usePyodide.tsx
@@ -40,17 +40,18 @@ export async function initPyodide({
 
   const status = loadStatusStore("python");
 
-  const indexURL = utils.currentScriptDir() + "/pyodide/";
+  const baseUrl = utils.currentScriptDir() + "/pyodide/";
 
   status.set("engine-download");
   // Checked because loadPyodide() hangs rather than failing when the wasm is
   // missing; see engine-load-guard.ts. This throw propagates to App.tsx, which
   // records it as "failed".
-  const unreachable = await checkEngineAssetReachable("python", indexURL);
+  const unreachable = await checkEngineAssetReachable("python", baseUrl);
   if (unreachable) throw new Error(unreachable);
 
   const pyodideProxy = await loadPyodideProxy(
-    { type: proxyType, indexURL },
+    // indexURL is Pyodide's own loadPyodide() parameter name; keep it as-is.
+    { type: proxyType, indexURL: baseUrl },
     stdout,
     stderr,
   );
--- a/src/hooks/usePyodide.tsx
+++ b/src/hooks/usePyodide.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect } from "react";
+import { checkEngineAssetReachable } from "../engine-load-guard";
 import { loadStatusStore } from "../load-status";
 import type { ProxyType, PyodideProxy } from "../pyodide-proxy";
 import { loadPyodideProxy } from "../pyodide-proxy";
@@ -39,12 +40,17 @@ export async function initPyodide({
 
   const status = loadStatusStore("python");
 
+  const indexURL = utils.currentScriptDir() + "/pyodide/";
+
   status.set("engine-download");
+  // Checked because loadPyodide() hangs rather than failing when the wasm is
+  // missing; see engine-load-guard.ts. This throw propagates to App.tsx, which
+  // records it as "failed".
+  const unreachable = await checkEngineAssetReachable("python", indexURL);
+  if (unreachable) throw new Error(unreachable);
+
   const pyodideProxy = await loadPyodideProxy(
-    {
-      type: proxyType,
-      indexURL: utils.currentScriptDir() + "/pyodide/",
-    },
+    { type: proxyType, indexURL },
     stdout,
     stderr,
   );
--- a/src/hooks/useWebR.tsx
+++ b/src/hooks/useWebR.tsx
@@ -133,8 +133,8 @@ export function useWebR({
       const webRProxyHandle = await webRProxyHandlePromise;
       setwebRProxyHandle(webRProxyHandle);
     })().catch((e) => {
-      // Already surfaced to the user via the load status store; log it so it
-      // isn't an unhandled rejection.
+      // Already surfaced to the user via the load status store;
+      // logged to the console so it isn't an unhandled rejection.
       console.error(e);
     });
   }, [webRProxyHandlePromise]);
@@ -346,28 +346,19 @@ webr::shim_install()
 # on to display an app that never started. Returning the failure as a value is
 # what makes a failed startup visible.
 #
-# A status list rather than the message alone: a condition whose message is
-# empty is indistinguishable from success once the two share one string, and
-# conditionMessage() on its own throws away the condition's call, so the viewer
-# could say what failed but not which call failed.
+# The status field carries the explicit outcome rather than leaving it to be inferred: a
+# condition can carry an empty message, which on its own would read as success.
+# The class and call fields are what conditionMessage() would drop, and the call
+# is how the dialog can name which call failed, not just what went wrong.
 .start_app <- function(appName, appDir, devMode = FALSE) {
   tryCatch(
     {
-      # Parse the app's R files before shiny does. shiny's sourceUTF8 catches the
-      # parse error and raises only "Error sourcing <file>", discarding the line
-      # number and the offending line, so letting parse() fail here is what gets
-      # that detail to the viewer. Runs first so a typo fails before any package
-      # installs rather than after them.
-      #
-      # Reporting the condition's call below does not replace this step: the
-      # detail is destroyed by sourceUTF8's own try() before any handler of ours
-      # sees a condition, so there is nothing left for one to report.
+      # Perform a basic parse of top-level R scripts to highlight any syntax
+      # errors. Runs first so a typo fails before spending time on package installs.
       for (f in list.files(appDir, pattern = "[.][Rr]$", full.names = TRUE)) {
         # call. = FALSE because the call this condition would otherwise carry is
         # this loop's own parse(file = f), whose f names nothing an app author
-        # would recognise. conditionMessage() keeps the whole reason the step
-        # exists: the file, the line and column, the offending source lines and
-        # the caret.
+        # would recognise.
         tryCatch(
           parse(file = f),
           error = function(cnd) stop(conditionMessage(cnd), call. = FALSE)
@@ -388,7 +379,7 @@ webr::shim_install()
         if (!has_pkg) {
           # Deliberately not fatal: renv::dependencies() also reports packages
           # that are named but never actually used, and those apps run fine
-          # today. A package that really is needed fails below, when the app
+          # today. A package that really is needed fails later, when the app
           # source is evaluated.
           webr::install(pkg_name)
         }
@@ -418,8 +409,7 @@ webr::shim_install()
         # shiny wraps every app body in ..stacktraceon..(), so for any top-level
         # failure the condition's call is the whole app source -- no use to anyone
         # reading the dialog. Report a call only when it names something the
-        # author would recognise, and fall back to the bare message otherwise,
-        # which is what the dialog showed before it reported a call at all.
+        # author would recognize, and fall back to the bare message otherwise.
         #
         # A prefix rather than a list of names: the shiny shipped here exports
         # ..stacktraceon.. and ..stacktraceoff.., and the prefix covers both
--- a/src/hooks/useWebR.tsx
+++ b/src/hooks/useWebR.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect } from "react";
 import { ChannelType } from "webr";
+import { loadStatusStore } from "../load-status";
 import * as utils from "../utils";
 import type { WebRProxy } from "../webr-proxy";
 import { loadWebRProxy } from "../webr-proxy";
@@ -38,6 +39,9 @@ export async function initWebR({
     : ChannelType.PostMessage;
   const baseUrl = utils.currentScriptDir() + "/webr/";
 
+  const status = loadStatusStore("r");
+
+  status.set("engine-download");
   const webRProxy = await loadWebRProxy(
     {
       baseUrl,
@@ -49,13 +53,16 @@ export async function initWebR({
 
   let initError = false;
   try {
+    status.set("engine-start");
     await webRProxy.webR.objs.globalEnv.bind(".base_url", baseUrl);
     await webRProxy.runRAsync(
       `webr::mount("/shinylive/library", "${baseUrl}library.data.gz")`,
     );
     await webRProxy.runRAsync(load_r_pre);
+    status.set("ready");
   } catch (e) {
     initError = true;
+    status.set("failed", e instanceof Error ? e.message : String(e));
     console.error(e);
   }
 
@@ -121,7 +128,11 @@ export function useWebR({
     (async () => {
       const webRProxyHandle = await webRProxyHandlePromise;
       setwebRProxyHandle(webRProxyHandle);
-    })();
+    })().catch((e) => {
+      // Already surfaced to the user via the load status store; log it so it
+      // isn't an unhandled rejection.
+      console.error(e);
+    });
   }, [webRProxyHandlePromise]);
 
   return webRProxyHandle;
--- a/src/hooks/useWebR.tsx
+++ b/src/hooks/useWebR.tsx
@@ -400,14 +400,23 @@ webr::shim_install()
       if (!nzchar(msg)) msg <- "The app failed to start, with no error message."
       # conditionCall() is NULL when the condition was signalled without a call,
       # and deparse() returns one element per line of source. Keep only the first
-      # line, marking that there was more, the way R's own error printing does:
-      # shiny evaluates the app's body inside ..stacktraceon..(), so deparsing
-      # that call in full would put the entire app source in the dialog.
+      # line, marking that there was more, the way R's own error printing does,
+      # so that a long but meaningful call is trimmed rather than dumped.
       cnd_call <- conditionCall(cnd)
       call_txt <- ""
       if (!is.null(cnd_call)) {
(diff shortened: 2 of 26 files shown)


jat255:
Local Claude review found a few issues worth fixing, and a couple "low" findings I think weren't worth the effort:

### Fixed

- **Python failing partway through startup hung the page forever** (`f6a7ec8`) — if the code that reports such errors crashed while composing the message, nothing was ever sent back and the loader span indefinitely. It now reports the error instead. With the failure forced, the old build sat on "Downloading Python…" with no error; the new one reports it in one.
- **A corrupted or substituted engine download wasn't detected** (`64651ed`) — the check only looked at the HTTP status, but a proxy sign-in page or a truncated file could also comes back as "200 OK". The bad file was then handed to the engine, which hangs rather than complaining. It now also confirms the response really is a WebAssembly file by checking the first few magic bytes of the file and cancelling the rest of the download. This was a bit of a nitpick from Claude, but probably worth fixing.
- **Cache-clearing advice appeared for ordinary app errors** (`6f99001`) — if a user's app had a typo or syntax error, the hard-refresh hint was shown which isn't actually useful in that case. The hard refresh advice now shows only when the engine itself fails to load.
- **A cosmetic startup step could kill a working app** (`da9365a`) — tidying the terminal and printing the startup banner were covered by the same error handling as engine startup, so a failure in either permanently replaced a running app with an error screen. Only steps that determine whether the engine actually works can do that now.

**Not addressed** — two minor points in the R engine code: a comment describing a failed package install as non-fatal when it actually stops startup (the comment is wrong, the behaviour is unchanged from `main`), and the syntax pre-check only looks at R files in the top-level folder.

src/hooks/usePyodide.tsx
```diff
@@ -39,12 +40,17 @@ export async function initPyodide({
 
   const status = loadStatusStore("python");
 
+  const indexURL = utils.currentScriptDir() + "/pyodide/";
```
cpsievert: Maybe let's call this `baseUrl` for consistency with the webr code paths?

src/hooks/useWebR.tsx
```diff
@@ -335,31 +339,50 @@ webr::shim_install()
...
+      }
 
-    if (!has_pkg) {
-      webr::install(pkg_name)
+      app <- .shiny_to_httpuv(appDir)
+      assign(appName, app, envir = .shiny_app_registry)
+      ""
+    },
+    error = function(cnd) {
+      msg <- paste(conditionMessage(cnd), collapse = "\n")
```
cpsievert, line 398: If the goal here is to capture precisely what the error would display as in the console, This might be too lossy of a thing that leads to the subsequent "`parse()` before eval" commit. Maybe this should return an actual error condition object instead?

src/pyodide-worker.ts
```diff
@@ -239,8 +247,13 @@ self.onmessage = async function (e: MessageEvent): Promise<void> {
       });
     }
   } catch (e) {
+    // Both of these can still be undefined here: init may have failed partway
+    // through, after loadPyodide() but before setupPythonEnv() returned. If we
+    // throw from this handler, no reply is ever posted and the caller waits
+    // forever, so it must not assume either one exists.
```
cpsievert: This feels a little bit too much like overexplaining the one particular line added here.

cpsievert (commented):
Here's a few things I noticed from just a quick sniff test. Also, I think the scope is large enough here that I would feel more comfortable shipping this if we had decent test coverage. Is that something you've looked into already?

jat255:
> Is that something you've looked into already?

We can (and I think we should) add some unit test coverage. I've been running a suite of tests locally (not included in the PR) for this set of changes in particular, but since the repo wasn't already setup for running unit tests, I didn't want to blow up the scope as part of this PR. Happy to do that either here or in a separate PR (probably more appropriate).

cpsievert:
Would it make sense to add to the `./playwright` tests?

jat255:
@cpsievert after talking with @schloerke and @karangattu, I'm going to make a separate PR to setup the unit test infrastructure with some baselines so this PR will have something to run against once getting rebased. I'll set it back to draft for now.

src/hooks/usePyodide.tsx
jat255: Renamed to `baseUrl` in `usePyodide.tsx`. One occurrence keeps the old name deliberately: `LoadPyodideConfig.indexURL` is `loadPyodide()`'s own parameter, and that config object is spread straight into the call, so renaming the field would break it. Added a comment there saying why the field name and the variable name differ.

src/hooks/useWebR.tsx
jat255: Two parts to this:

1. **It now carries the condition's fields, rather than a bare string (and still not the whole condition).**  A whole condition can't cross the boundary usefully: `conditionCall()` is a language object, and webR's `toJs()` throws on those, so the one part worth having has to be deparsed in R either way. `.start_app` now returns `list(status = "ok")` or `list(status = "error", message, class, call)`, instead of `""`-for-success and the message for failure. That old scheme couldn't tell success apart from an error condition whose message happened to be empty. It appears this smell was already known, since the code patched around the collision with a "no error message" fallback. `Viewer.tsx` now reads it with `evalR` + `toJs()` on the shelter it already opens.

   `call` is the new useful part for a user: R prints `Error in library(nonexistent): there is no package called 'nonexistent'`, and the dialog previously showed only the half after the colon. One caveat — shiny wraps every app body in `..stacktraceon..()`, so for a *top-level* failure `conditionCall()` deparses to the entire app source. That wrapper is suppressed, so a call is reported only when it names the author's own code. `class` is carried as well, but only as far as the console; giving the error screen a third category felt like a separate change.

2. **The `parse()` loop is a separate concern, so I've left it in.** It runs over the app's top-level `.R` files *before shiny sees them* because Shiny's `sourceUTF8` catches a parse error with `try()` and re-raises a bare `Error sourcing <file>`. By the time any handler of ours runs — either the old string format or the new condition — the information about line and caret position is already gone. Parsing first is what gets `app.R:6:0: unexpected end of input` and the caret into the dialog, and it would still be needed under any return shape. I've added that reason to its comment for clarity.

   Regardless, there's no traceback: `tryCatch` unwinds before the handler runs, so `sys.calls()` there is already empty, and doing it properly needs `withCallingHandlers` in the path that starts every R app. This would be a more invasive change, like we talked about.

src/pyodide-worker.ts
jat255: Trimmed to two lines — kept the part that is not obvious from the code (`pyUtils` can still be undefined here if init failed before `setupPythonEnv()` returned) and dropped the background about the reply never being posted, which the surrounding try/catch already shows.

jat255:
@cpsievert this should be ready for re-review.

The PR is now covered using the infrastructure setup in #246 and #247.

Added **37 jest unit tests** covering the pure logic: `src/load-status.test.ts` for the status store (including the subscription semantics `useSyncExternalStore` depends on, per-engine isolation, `failed` being terminal) and `src/engine-load-guard.test.ts` for the reachability and wasm-magic-number checks, including a magic number split across chunks, a body that cannot be cancelled, and one that throws mid-read. 100% on all four coverage metrics over the modules they import (this PR didn't get the coverage comment because it's from my fork -- a vestige from before I had write access):

<details>
<summary><b>Coverage output (100%)</b></summary>
<pre><code>
npm run test:unit:coverage

> shinylive@0.10.14 test:unit:coverage
> node testing-helpers/coverage-note.js

 PASS  src/Components/codeMirror/language-server/diagnostics.test.ts
 PASS  src/parse-codeblock.test.ts
 PASS  src/messageportwebsocket.test.ts
 PASS  src/Components/ResizableGrid/DragToResizeHelpers.test.ts
 PASS  src/r-status.test.ts
 PASS  src/engine-load-guard.test.ts
 PASS  src/load-status.test.ts
 PASS  src/Components/share.test.ts
 PASS  src/fileio.test.ts
 PASS  src/utils.test.ts
 PASS  src/examples.test.ts
 PASS  src/Components/codeMirror/utils.test.ts
 PASS  src/Components/codeMirror/language-server/positions.test.ts
 PASS  src/postable-error.test.ts
 PASS  src/Components/gist.test.ts
 PASS  src/Components/utils.test.ts
 PASS  src/Components/filecontent.test.ts
 PASS  src/awaitable-queue.test.ts
 PASS  src/Components/codeMirror/language-server/regexp-util.test.ts
 PASS  src/Components/codeMirror/language-server/names.test.ts
-------------------------------------------|---------|----------|---------|---------|-------------------
File                                       | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-------------------------------------------|---------|----------|---------|---------|-------------------
All files                                  |     100 |      100 |     100 |     100 |                   
 src                                       |     100 |      100 |     100 |     100 |                   
  awaitable-queue.ts                       |     100 |      100 |     100 |     100 |                   
  engine-load-guard.ts                     |     100 |      100 |     100 |     100 |                   
  examples.ts                              |     100 |      100 |     100 |     100 |                   
  fileio.ts                                |     100 |      100 |     100 |     100 |                   
  load-status.ts                           |     100 |      100 |     100 |     100 |                   
  messageportwebsocket.ts                  |     100 |      100 |     100 |     100 |                   
  parse-codeblock.ts                       |     100 |      100 |     100 |     100 |                   
  postable-error.ts                        |     100 |      100 |     100 |     100 |                   
  r-status.ts                              |     100 |      100 |     100 |     100 |                   
  utils.ts                                 |     100 |      100 |     100 |     100 |                   
 src/Components                            |     100 |      100 |     100 |     100 |                   
  filecontent.ts                           |     100 |      100 |     100 |     100 |                   
  gist.ts                                  |     100 |      100 |     100 |     100 |                   
  share.ts                                 |     100 |      100 |     100 |     100 |                   
  utils.ts                                 |     100 |      100 |     100 |     100 |                   
 src/Components/ResizableGrid              |     100 |      100 |     100 |     100 |                   
  DragToResizeHelpers.ts                   |     100 |      100 |     100 |     100 |                   
 src/Components/codeMirror                 |     100 |      100 |     100 |     100 |                   
  utils.ts                                 |     100 |      100 |     100 |     100 |                   
 src/Components/codeMirror/language-server |     100 |      100 |     100 |     100 |                   
  diagnostics.ts                           |     100 |      100 |     100 |     100 |                   
  names.ts                                 |     100 |      100 |     100 |     100 |                   
  positions.ts                             |     100 |      100 |     100 |     100 |                   
  regexp-util.ts                           |     100 |      100 |     100 |     100 |                   
-------------------------------------------|---------|----------|---------|---------|-------------------

Test Suites: 20 passed, 20 total
Tests:       342 passed, 342 total
Snapshots:   0 total
Time:        3.394 s
Ran all test suites.
</code></pre></details>

Added **19 pytest + Playwright tests** based off how I was manually testing using the demo script (`tests/test_loader_status.py`, `-m loader`) driving the failure modes end to end against the built site. These cover engine asset 404, connection dropped for that asset, a post-boot `setupPythonEnv()` failure, app syntax errors in both engines, an unresolvable Python requirement (fatal) versus an unresolvable R library (non-fatal by design, app still runs), the R condition's message and call reaching the dialog, the staged loading text on a slow load, and the embedded-block error screen. Exercises both Python and R engines, in both full screen and embed layouts.

One thing to flag: *The new tests don't gate the deploy.* They run in their own workflow, `test-loader.yml`, following the pattern of `test-apps.yml`. The other half of the `site` marker (the old tests re-enabled in #247) stays in `build.yml` and still stops the deploy. Moving the gate later is either folding the marker back into `make test-site`, or a `workflow_run` trigger on this workflow's success.

For reference,`tests/README.md` documents how to watch or record a scenario through Playwright.

cpsievert (approved):
Nice work, thanks!

PR #241, opened 2026-08-07
test: strengthen example intent coverage

karangattu:
## Summary

This change detects errors that occur after an intent test interacts with an example. A shared fixture checks terminal output, Shiny output errors, browser console errors, and page errors.

The Python intent tests now confirm these results:

- CPU Info updates the heatmap after a colormap change.
- Regularization redraws all three dependent plots.
- Altair redraws its chart after a variable change.
- Map moves its viewport after a city change.
- Fetch renders a mocked API response without an external request.

### DIFF
--- a/tests/test_examples_smoke.py
+++ b/tests/test_examples_smoke.py
@@ -18,4 +18,5 @@ EXAMPLES = [(engine, title) for engine in ENGINES for title in example_app_title
     "engine,title", EXAMPLES, ids=[f"{engine}-{title}" for engine, title in EXAMPLES]
 )
 def test_example_app_starts_cleanly(page: Page, engine: str, title: str) -> None:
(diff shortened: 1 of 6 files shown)


tests/test_examples_smoke.py
```diff
@@ -26,28 +18,4 @@
...
-            return
-        # A missing favicon is not an app problem.
-        if message.location["url"].endswith("favicon.ico"):
-            return
-        console_errors.append(message.text)
-
-    page.on("console", on_console)
-    page.on("pageerror", lambda error: console_errors.append(str(error)))
-
     open_example(page, engine, title)
```
schloerke, line 22:
```suggestion
    # fail_on_example_errors fixture (autouse=true) will test for console errors.
    open_example(page, engine, title)
```

schloerke (approved):

Issue #243, opened 2026-08-11
Get orphaned Playwright specs running again and setup infrastructure for future tests


jat255:
## Summary

`playwright/` has four specs that predate the examples suite. `make test` calls a script that doesn't exist, and two of the four need `shinylive export` from the Python `shinylive` package, which nothing here installs. #236 hit this same wall and deliberately routed around it rather than fixing it, and #238 has since landed and moved app testing to pytest.

The goal here is to **get all four running again**, not abandon them — they cover ground the examples suite doesn't (the editor, loading apps from URL parameters, static-export output). The how is mostly settled by the house style #238 established: they migrate to pytest under `tests/`. The real work is a local export fixture that breaks the Python-package dependency for the two static-export specs.

To be clear about what already works: **the example-app suite runs in CI** via `test-apps.yml`. #236 added it as a Playwright Test suite; #238 migrated it to pytest under `tests/` (deleting `playwright-examples.config.ts`) and added intent tests alongside the smoke tests. That suite is healthy. This issue is only about the older specs it stepped around.

Companion issue: #242, fixing the jest harness, which is clear-cut and landing first as its own PR.

One scheduling constraint: *#240 (loading screen) will ship with its own Playwright tests included*. Those are app-level tests for non-example apps — exactly the ground this issue covers — so whatever here is a prerequisite for them needs to land first. See Sequencing.

## What's there

`playwright/` now holds only the four specs that predate the examples suite — #238 took the smoke specs with it when it moved to `tests/`:

| Spec | Server | Needs the Python `shinylive` package? |
|---|---|---|
| `examples-viewer.spec.ts` | `:3000` (`npm run test-server`) | No |
| `load-from-url.spec.ts` | `:3000` (`npm run test-server`) | No |
| `shiny-static.spec.ts` | `:8008` (`shinylive export`) | Yes |
| `editor-cell.spec.ts` | `:8009` (`shinylive export`) | Yes |

Two problems keep all four from running.

**`make test` calls a script that doesn't exist.**

```make
test:                                    # Makefile:389
	npm run playwright test
```

The scripts in `package.json` are `serve`, `test-server` and `cypress:open`. There's no `playwright` script, so this just fails. It should be `npx playwright test`. (`cypress:open` is dead too — `cypress` isn't a dependency.)

[...]

PR #244, opened 2026-08-11
test: Fix the jest harness and add a baseline unit suite

### LINKED ISSUE #242: Fix/add jest harness and unit tests
## Summary

There's jest config in `package.json` and two committed test files, but the transform it needs and the mock directory it points at were never added. So there's currently no way to write a unit test for anything in `src/`. This isn't bit-rot — as far as I can tell the harness has never run at any commit.

Discussed in person with @schloerke and @karangattu (2026-08-11), and the direction is agreed:

- **Jest is fine for TypeScript unit tests** — no objections. Use it for pure logic; DOM/store interaction likely belongs in Playwright instead, with Testing Library (already in `devDependencies`) as an option for DOM mocking.
- **This lands first** as a no-logic-changes PR (harness + CI wiring + baseline tests), so the loading screen PR (#240) — which now ships with its own Playwright tests included — has existing sanity tests it must pass before it merges.

Companion issue: #243, getting the four orphaned Playwright specs running again — same neighborhood, but that one is migration work and this one isn't.

## What's there now

`package.json` has a `jest` key:

```jsonc
"jest": {
  "testEnvironment": "jsdom",
  "modulePathIgnorePatterns": ["pyodide", "testing-helpers/skeleton.test.tsx"],
  "moduleNameMapper": {
    "\\.(css|less|sass|scss)$": "<rootDir>/testing-helpers/__mocks__/styleMock.js",
    "\\.(gif|ttf|eot|svg)$": "<rootDir>/testing-helpers/__mocks__/fileMock.js"
  }
}
```

and the dependencies to go with it:

| Package | Version |
|---|---|
| `jest` | `^27.5.1` |
| `@types/jest` | `^29.5.11` (doesn't match jest 27) |
| `@testing-library/react` | `^14.1.2` |
| `@testing-library/dom` | `^8.20.1` |
| `@testing-library/user-event` | `^14.5.1` |

Two test files are committed: `src/Components/Editor.test.tsx` and `src/Components/ExampleSelector.test.tsx`. Running either one fails right away:

```
FAIL src/Components/ExampleSelector.test.tsx
  ● Test suite failed to run
    Jest encountered an unexpected token
    Jest failed to parse a file...
```

## This isn't bit-rot, it never worked

| When | Commit | What happened |
|---|---|---|
| 2022-05-17 | `9fdb3a1` "Move files from srcjs/ to /" | jest config, both test files, `babel-jest` and four `@babel/preset-*` all arrive together |
| 2022-05-26 | `500c6c0` "Update more node modules and remove unused" | removes `babel-jest` and all four presets |

`500c6c0` looks like the culprit, but it was right to remove them. Three things have never existed in this repo at any commit:

- **A babel config.** No `.babelrc`, no `babel.config.*`, and `git log -S'"babel"' -- package.json` comes back empty, so there was never a `"babel"` key either. Without a config those presets did nothing, which is exactly what that commit message says.
- **The `testing-helpers/` directory.** `git log --all -- '*testing-helpers*'` is empty, but `moduleNameMapper` and `modulePathIgnorePatterns` both point into it.
- **A jest `test` script.** At `9fdb3a1` the scripts were `watch` and `build-react`. `make test` has only ever run Playwright.

It looks like the config was copied in from wherever `srcjs/` came from, without the files it depends on.

The reason you get a parser error rather than "no transform configured" is that `babel-jest@27.5.1` is still installed transitively as jest's default transform. It runs, but with no presets, so `@babel/parser` hits the first bit of TypeScript and throws.

## Work

- [ ] Add a transform. `@swc/jest` seems like the better pick over `ts-jest`: nothing to configure and no type-checking in the test path, which `tsc --noEmit` already handles separately.
- [ ] Add the missing `testing-helpers/__mocks__/styleMock.js` and `fileMock.js`, or point `moduleNameMapper` somewhere that exists. The `testing-helpers/skeleton.test.tsx` entry in `modulePathIgnorePatterns` also refers to a file that was never added.
- [ ] Add `testPathIgnorePatterns` for `/playwright/`. Right now `npx jest --listTests` picks up all four Playwright specs and would try to run them.
- [ ] Add a `test:unit` script.
- [ ] Sort out `@types/jest` vs jest 27, or bump jest. If you do bump, note that jest 28+ doesn't bundle `jest-environment-jsdom` any more, so it needs installing alongside the `testEnvironment: "jsdom"` setting.
- [ ] Decide what to do with the two existing test files. They were edited at `f0e9eea` and `5d8c287` without ever being run, so they'll probably fail on content as well as config. Worth fixing or deleting on purpose rather than leaving them broken.
- [ ] Generate a baseline of unit tests across the existing `src/` so the harness lands with real coverage rather than an empty runner. This was Barret's suggestion, and LLM assistance makes it much cheaper than it used to be. Pure logic first; anything touching the DOM or the store is probably better served by Playwright than jsdom.
- [ ] Wire `test:unit` into CI so a failure blocks merge.

## Acceptance criteria

- [ ] `npm run test:unit` runs jest from a clean clone after `npm install`, with no references to files that don't exist, and a jest failure in CI blocks the merge.
- [ ] A short note in the README or CONTRIBUTING covering which suites exist and how to run each — jest for unit tests, pytest (via the `tests/` directory and its Make targets) for app-level browser tests.

## Out of scope

- Adding feature coverage beyond the generated baseline. This is about making the harness work so tests can come after.
- The four orphaned Playwright specs in `playwright/` — that's #243. (One exception: `testPathIgnorePatterns` for `/playwright/` is in scope here, since jest would otherwise try to run them.)
- Loading screen test authoring — those tests ship as part of #240 itself (see Sequencing), though the harness work here is what unblocks them.
- Any change to `src/`.

## Sequencing

Order from the 2026-08-11 discussion, with one update: the loading screen Playwright tests now ship **as part of #240**, not as a follow-up PR.

1. ~~**#238 (Karan's example-app suite) merges first.**~~ **Done** — merged 2026-08-07. Smoke and intent tests for all 38 example apps run via pytest in `tests/`, serving the local build at `127.0.0.1` to route around the circular dependency.
2. **This issue lands as its own PR** — Jest, CI wiring, and generated baseline unit tests, with no logic changes. Self-contained and carries no CI risk.
3. **Any prerequisite work from #243 lands.** #240's loading screen tests are the first app-level tests for non-example apps, so they need a settled home and possibly a shared serving fixture; see #243's Sequencing section for what that does and doesn't require.
4. **#240 (loading screen) merges last.** It must pass the sanity tests from step 2, and it includes its own deterministic Playwright tests — happy path, error states, and stage transitions — run via pytest in CI like everything else.

## Why now

PR #240 (loader status and failure reporting) adds `src/load-status.ts` and `src/engine-load-guard.ts`. Both are pure and dependency-free and would be easy to unit test, and the PR adds several error paths — engine/WASM download failure, app syntax errors, stage transitions — that would make reasonable browser tests. Those tests were written and run locally, but there's nowhere to commit them where they'd actually be executed. Fixing this issue alone would unblock most of that.

Two broader points from the 2026-08-11 discussion:

- Generally speaking in Shinylive, holistic integration tests matter more than line coverage. The baseline unit tests are a floor, not the goal — the pytest-driven app suites are what actually protect the repo.
- The repo has been historically undertouched largely because the missing test harness kept contributor confidence low. Most of the code in this area traces back ~4 years to well-established patterns, so a generated baseline is capturing stable behavior, not chasing a moving target.

## References

- `package.json` — the `jest` key, test dependencies, `scripts`
- `9fdb3a1`, `500c6c0` — where the jest config came from, and when the presets went
- #240 — the loading screen PR this unblocks
- #232 / #238 — the move to pytest for app-level tests (merged 2026-08-07), which is why DOM/store testing stays out of jest's scope

jat255:
Closes #242.

The `jest` key in `package.json` has never been able to run. As #242 works out, this isn't bit-rot: the transform it needs, the `testing-helpers/` directory `moduleNameMapper` points at, and a `test` script have never existed at any commit. The two committed test files were edited twice without ever being executed.

## Harness

`@swc/jest` as the transform, over `ts-jest` — nothing to configure, and no type-checking in the test path, which `tsc --noEmit` handles separately. jest goes 27 → 29 to match the `@types/jest` already pinned at `^29`, which means `jest-environment-jsdom` has to be installed alongside it since jest stopped bundling it in 28.

The config moves out of `package.json` and into `jest.config.js`. Given that the old config quietly misled everyone for four years, the non-obvious parts are worth a comment, and JSON can't carry one.

`testing-helpers/` now exists — the two asset mocks `moduleNameMapper` has always wanted, plus a setup file for browser globals jsdom doesn't implement (`TextEncoder`, `structuredClone`; the latter matters because `postable-error.ts` exists specifically to be structured-cloneable). The `skeleton.test.tsx` entry in `modulePathIgnorePatterns` is gone, and the bare `"pyodide"` pattern is scoped to `<rootDir>/src/pyodide`. `roots` is `src/` and `testPathIgnorePatterns` covers `playwright/`, so `jest --listTests` collects exactly the 16 unit test files and none of the Playwright specs.

One thing not on the issue's checklist: `testEnvironmentOptions.customExportConditions: [""]`. jest-environment-jsdom resolves package `exports` with the `browser` condition, which sends `vscode-languageserver-protocol` to an ESM build the CJS loader can't parse, and the LSP tests fail to load without it. Both this and the jsdom polyfills were checked to be load-bearing by removing them and watching tests fail, rather than left in on suspicion.

`tsconfig.json` gains `"types": ["jest"]` and stops excluding test files, so `tsc --noEmit` type-checks them. A separate `tsconfig.test.json` would be tidier — it would keep jest's globals out of the app's type environment — but the test files reach `src/Components/App.tsx` through a type-only import, which drags in `pyodide-proxy.ts` and its pre-existing error, so that config could never be green.

## The two existing test files

Deleted rather than fixed. Neither can be made to work without changing `src/`, which #242 puts out of scope:

- `ExampleSelector.test.tsx` imports `../examples.json`, which does not exist — examples are fetched at runtime from `../shinylive/examples.json` — and default-imports a named export, and omits the now-required `appEngine` prop.
- `Editor.test.tsx` imports `babel-polyfill` and a `testing-helpers/clearThenType`, neither of which has ever existed at any commit, and passes three props to an `Editor` that now requires eight.

Both are component tests of the kind the 2026-08-11 discussion moved to Playwright and pytest, so nothing that belongs in jest is lost.

## Tests

228 tests across 16 files, all pure logic: path and filetype helpers, binary detection and byte conversion, the codeblock parser, file content encoding, share URLs, gist decoding, the File System Access wrappers, the MessagePort WebSocket state machine, and the codemirror/LSP position and diagnostic mapping. Anything needing a real browser, a running app, or the engines is left to the app tests in `tests/`.

The suite was mutation-tested rather than trusted — each source file was deliberately broken to check the tests noticed. That pass caught several that would have passed against a broken implementation, and they've been tightened: the "send() after close is dropped" test passed with the guard deleted, the UTF-8 tests were ASCII-only, `sleep()` never checked its argument, and the `structuredClone` assertion could not fail. There is also now a test for the late-`close`-frame path, which nothing else pinned.

Five latent bugs in `src/` turned up. All are left alone under the no-`src`-changes rule, and each is pinned by a characterization test whose comment says it documents the behavior rather than endorses it:

- **`parseCodeBlock()` crashes on a codeblock with no `#|` args.** `processQuartoArgs()` returns `undefined` for `quartoArgs` — js-yaml's result for an empty document — and `parseCodeBlock()` dereferences it, so the user gets a `TypeError` instead of the intended "must have a '#| standalone: true' argument". Probably worth its own issue.
- **`loadDirectoryRecursive()` forwards neither `maxBytes` nor `maxFiles` to subdirectories**, so a subdirectory silently reverts to the 20 MB / 20 file defaults.
- **Its file-count guard runs before the push**, so `maxFiles` admits `maxFiles + 1` files.
- **`errorToPostableErrorObject(null)` and `postableErrorObjectToError(null)` both throw `TypeError`**, and both callers are `catch (e)` blocks, where any value can arrive.
- **`names.ts` on a string with no `(`:** `nameFromSignature("foo")` returns `""` and `removeFullyQualifiedName("a.b.c")` returns its input with the qualifier unstripped, both from `substring(0, -1)`. Unreachable from current callers, which always pass a real signature.

Two smaller things, not bugs: the `newOffset > cmDoc.length` branch in `codeMirror/utils.ts` is unreachable, and the two arms of `saveFileContentToFile`'s `if (file.type === "binary")` are byte-identical.

## CI and docs

A `test_unit` job in `build.yml` with its own minimal checkout, so it doesn't wait on the full build. The `build` job gates on it, since `build` is also what deploys — without that, a failing unit test couldn't stop a deploy. Making a jest failure block a *merge* additionally needs branch protection to list "Unit tests" as a required check, which isn't something this PR can do.

A `## Testing` section in the README covers both suites and points at `tests/README.md`. It also notes that `make test` currently calls an `npm run playwright` script that doesn't exist — #243 removes it.

## Verification

`npm run test:unit` after a clean `npm ci` in a fresh worktree: 16 suites, 228 tests, all passing, ~0.5s. `npm ci --dry-run` reports the lockfile consistent, and only the jest/jsdom/babel/swc toolchain moved in it.

`npx tsc --noEmit` reports 7 errors, all pre-existing on `main` — verified by running it on a clean `fe8c0b6` worktree and getting a byte-identical list (`src/pyodide-proxy.ts:592` and six in `src/scripts/codeblock-to-json.ts`). None come from the test files. Because of those 7, `tsc --noEmit` can't go into CI, so type errors in tests are still only caught by hand.

The new test files add zero eslint problems; the one remaining error, in `webr-proxy.ts`, is pre-existing on `main`.

## Not included

`@testing-library/*` stays in devDependencies unused, since #242 names it as the option for future DOM mocking. `moduleNameMapper`'s CSS and asset stubs are likewise unexercised — no current test imports either — but they're what the config has always pointed at. `playwright/`, `playwright.config.ts`, the `Makefile` `test` target and `tests/` are untouched; that's #243.


jat255:
Superseded by #246 — same commits, from a branch on posit-dev rather than a fork.

PR #246, opened 2026-08-11
test: Fix the jest harness and add a baseline unit suite

### LINKED ISSUE #242: Fix/add jest harness and unit tests

jat255:
Closes #242.

Adds a working unit test harness and tests for everything that can be unit tested (logic that doesn't rely on React components or the DOM).

One deliberate departure from #242: it asks for no `src/` changes, but this PR includes four: three bug fixes that the new tests and the type-check discovered, and comment-only pragmas excluding three unreachable lines from coverage. See below.

## Harness

[`jest.config.js`](https://github.com/posit-dev/shinylive/pull/246/files#diff-1e058ca1442e46581b13571fb8d261f9e1f5657e26c96634d4c1072f0f0347f1) — config moves out of package.json so the non-obvious parts can carry a comment. Uses @swc/jest over ts-jest (no config, and tsc --noEmit already type-checks); jest updated from v27 → v29 to match the @types/jest already pinned, which means installing jest-environment-jsdom separately. Two workarounds are commented in place: customExportConditions and the jsdom globals in testing-helpers/jest.setup.js. Both were confirmed load-bearing by removing them and watching tests fail.

[`Makefile`](https://github.com/posit-dev/shinylive/pull/246/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52) — `make unit-test` and `make unit-test-coverage`, named for the `examples-*-test` pattern already in use rather than mirroring the npm script. Both depend on the existing `node_modules` target, so a checkout with a stale lockfile installs first. The npm scripts stay for the watch loop.

`@testing-library/*` is removed — three direct packages and sixteen transitive ones. Its only importers were the two test files deleted below, so nothing in the repo referenced it. #242 names it as an option for DOM mocking later, but installing it when something actually needs it costs one command, and would get a coherent pair: the pinned set never was one, since `@testing-library/react@14` expects `@testing-library/dom@^9` and the manifest asked for `^8`.

[`tsconfig.json`](https://github.com/posit-dev/shinylive/pull/246/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0) — stops excluding test files, so `tsc --noEmit` covers them. A separate `tsconfig.test.json` would be tidier but can't be green: the test files reach `App.tsx` by a type-only import, which drags in `pyodide-proxy.ts` and its pre-existing error.

## The two existing test files

This PR [deleted](https://github.com/posit-dev/shinylive/pull/246/files#diff-6ed75e3a102f51f1b598ffd25a76e6b96b931c4a9bb2294886c86f255e025cd9) the two existing test files. Both were phantom tests that imported modules that have never existed at any commit (`babel-polyfill`, `testing-helpers/clearThenType`, `../examples.json`), so neither can be repaired without changing `src/`. They're component tests of the kind the 2026-08-11 discussion moved to pytest anyway.

## Tests

Added 291 tests in 17 files; taking ~3s to run — pure logic only. Each source file was deliberately broken to check the tests noticed; the ones that survived have been tightened.

The largest single addition is [`DragToResizeHelpers.ts`](https://github.com/posit-dev/shinylive/pull/246/files#diff-10d190e16bce3c2e57c4ca0bb83afec693edd096b7a7ed4c0d539b6df02f529a) — 344 lines of grid-track arithmetic that had no tests and needs no browser, only a stubbed `getComputedStyle`.

This process revealed eight latent `src/` bugs. Three of them are error paths that threw over the error they were meant to report, and those are fixed here:

- **`parseCodeBlock()`** dereferenced `yamlLoad()`'s result unguarded. An empty YAML document — a block with no `#|` lines — gives `undefined`, so the likeliest authoring mistake raised a `TypeError` rather than the `'#| standalone: true'` message written for that case. Reachable from `run-python-blocks.ts` and the `codeblock-to-json` CLI.
- **`to_html`** declared `_to_html` as returning a `ToHtmlResult`, when a Python function hands back a `PyProxy` — which is what the `toJs()` call converts. The wrong type hid a fallback that returned a plain object with no `toJs()`, so the branch meant to report a missing `_to_html` threw instead and lost the message. Found by turning the type-check on.
- **`errorToPostableErrorObject()`** read `e.name` before its `instanceof Error` guard, and **`postableErrorObjectToError()`** used `in` on an unchecked value. The former's only caller is a `catch` in `pyodide-worker.ts`, so `throw null` anywhere in the worker made the reporter throw and lose the original error.

The remaining five are left in place. None is reachable today — no current caller passes the input that would trigger it — so fixing them would mean changing `src/` for no observable gain. Each instead has a test that locks in the behaviour as it stands, commented to say it records the bug rather than blesses it, so that changing the code later trips the test and surfaces this decision:

- [`loadDirectoryRecursive()`](https://github.com/posit-dev/shinylive/pull/246/files#diff-ac5e759b20e993b8ac1c9577cce4ae3faa3e058d76282edf61b35c7f00fec7c8) doesn't pass `maxBytes` or `maxFiles` down when it recurses, so a subdirectory silently reverts to the 20 MB / 20 file defaults.
- The same function checks the file count *before* adding the file, so it admits one more than `maxFiles`. Deserves its own issue rather than a quick fix: the limit is per-directory rather than a total for the tree, so making it behave "correctly" means first deciding what it ought to mean.
- `getExampleCategories()` caches its result but doesn't key the cache by engine, so asking for R after Python hands back Python's examples.
- Both `fileContentsToUrlString` variants call `.sort()` on the array they are given, reordering the caller's array as a side effect.
- `names.ts` mishandles a string containing no `(`: `substring(0, -1)` leaves the qualifier unstripped instead of failing.

## CI and coverage

[`build.yml`](https://github.com/posit-dev/shinylive/pull/246/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721) — added a `test_unit` job that checks out and installs on its own, so it doesn't sit behind the full build. It runs `make type-check` as well as the tests: swc strips types without checking them, so without that step nothing checks them at all.

The `build` job declares `needs: test_unit`. That matters because `build` doesn't only build, it also deploys to S3 on the `deploy` branch, uploads the `_shinylive/` artifact on `main`, and attaches assets on a release. Without the dependency those steps would run regardless of whether the unit tests passed. The gate costs the `build` step almost nothing, since `test_unit` finishes in well under a minute.

On a PR, coverage is measured with per-file deltas against the base and annotations on uncovered lines, then posted as a sticky comment via `marocchino/sticky-pull-request-comment`, so a busy PR gets one comment updated in place rather than one per push. That action was picked over the hand-rolled `gh api` script this PR originally used, per review: it is already used in ~14 workflows across ~10 posit-dev repos, and the script had a latent bug — piping `--paginate` into `head -1` exits 141 under `pipefail` once a PR has more than one page of comments. Both third-party actions hold a write token, so both are pinned to a commit SHA. The report also goes to the run's summary page, which survives if commenting fails.

Coverage is reported but never enforced — there is no threshold. A PR from a fork only gets a read-only token, so it can do neither the annotations nor the comment; it skips those steps through an explicit `if` rather than running them under `continue-on-error`, which keeps the skip visible in the UI and any genuine failure loud.

Two coverage settings carry their reasoning as comments in `jest.config.js`. `coverageProvider: "v8"` is used rather than jest's default instruments through `babel-jest`.  This config doesn't use `babel-jest`, so `--coverage` would have reported nothing at all. There is also deliberately no `collectCoverageFrom`, because widening collection to every file in `src/` would include all the files that cannot be unit tested and show coverage as ~15%. This setup calculates coverage based on what *can be unit tested*.

Coverage is at 100% on all four metrics, but only over the modules these tests import, so [the run prints a note](https://github.com/posit-dev/shinylive/pull/246/files#diff-aa599280617d537e92a2ae7707922363e79be63e6f7fe102b4da8d43d809dbed) under the table and the CI comment carries the same caveat. The rest of `src/` (~7,300 of ~8,700 statements: React, hooks, engine proxies, service worker) is the pytest suites' job and isn't measured here. Three unreachable lines are explicitly ignored in the coverage calc.

**Ignore the red deltas on this PR** — the base for this PR is the current `main` branch, which doesn't have jest properly configured, so the action assumes a 100% baseline that never existed. Fixed once this merges.

## Verification

`make unit-test` from a clean install: 17 suites, 291 tests, passing in < 1s. CI is green, including the coverage comment above.

`make unit-test-coverage` from a clean install: same stats, takes about 3s. 

Coverage report:
```
-------------------------------------------|---------|----------|---------|---------|-------------------
File                                       | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-------------------------------------------|---------|----------|---------|---------|-------------------
All files                                  |     100 |      100 |     100 |     100 |                   
 src                                       |     100 |      100 |     100 |     100 |                   
  awaitable-queue.ts                       |     100 |      100 |     100 |     100 |                   
  examples.ts                              |     100 |      100 |     100 |     100 |                   
  fileio.ts                                |     100 |      100 |     100 |     100 |                   
  messageportwebsocket.ts                  |     100 |      100 |     100 |     100 |                   
  parse-codeblock.ts                       |     100 |      100 |     100 |     100 |                   
  postable-error.ts                        |     100 |      100 |     100 |     100 |                   
  utils.ts                                 |     100 |      100 |     100 |     100 |                   
 src/Components                            |     100 |      100 |     100 |     100 |                   
  filecontent.ts                           |     100 |      100 |     100 |     100 |                   
  gist.ts                                  |     100 |      100 |     100 |     100 |                   
  share.ts                                 |     100 |      100 |     100 |     100 |                   
  utils.ts                                 |     100 |      100 |     100 |     100 |                   
 src/Components/ResizableGrid              |     100 |      100 |     100 |     100 |                   
  DragToResizeHelpers.ts                   |     100 |      100 |     100 |     100 |                   
 src/Components/codeMirror                 |     100 |      100 |     100 |     100 |                   
  utils.ts                                 |     100 |      100 |     100 |     100 |                   
 src/Components/codeMirror/language-server |     100 |      100 |     100 |     100 |                   
  diagnostics.ts                           |     100 |      100 |     100 |     100 |                   
  names.ts                                 |     100 |      100 |     100 |     100 |                   
  positions.ts                             |     100 |      100 |     100 |     100 |                   
  regexp-util.ts                           |     100 |      100 |     100 |     100 |                   
-------------------------------------------|---------|----------|---------|---------|-------------------

Test Suites: 17 passed, 17 total
Tests:       291 passed, 291 total
Snapshots:   0 total
Time:        2.997 s
Ran all test suites.

Unit test coverage only. This covers the pure-logic modules these
tests import -- parsing, encoding, paths, grid maths and the like.

It excludes most of src/: the React components, the hooks, the engine proxies
and the service worker. Those need a real browser and a running app, and are
covered by the pytest + Playwright suites in tests/, whose coverage is not
measured here. A figure near 100% above does not mean shinylive is fully
tested.

See the Coverage section of README.md.
```

`make type-check` is clean. It reported 7 errors on `main`, which is why it had never been wired up; those are fixed here. Six were `src/scripts/codeblock-to-json.ts`, a Deno script that imports over https and uses the `Deno` global, so a browser tsconfig can never resolve it — it is excluded now, and since esbuild takes it as an explicit entry point the bundle is unchanged. The seventh was the `to_html` bug above.

## Not included

`playwright/`, `playwright.config.ts` and the broken `test` target are untouched — that's #243 / #247.

Expect a small Makefile conflict with #247, which edits the same region: it drops the `test` target, renames `examples-test-deps` to `test-deps`, and adds `site-test`. Whichever lands second resolves it there and in `.PHONY`.


.github/workflows/build.yml
```diff
@@ -12,7 +12,80 @@ on:
...
+      contents: read
+      pull-requests: write
+      checks: write # uncovered-line annotations
+
+    steps:
+      - uses: actions/checkout@v7
+
+      - uses: actions/setup-node@v7
+        with:
+          node-version: "20.x"
```
schloerke: Unless precedent in repo, this should be `lts/*`

schloerke: I don't see one, so let's use the lts version (22 or 24)
```suggestion
          node-version: "lts/*"
```

.github/workflows/build.yml
```diff
@@ -12,7 +12,80 @@ on:
...
+
+      - name: Run unit tests
+        run: npm run test:unit
+
+      # Coverage is not a gate, just informative for PRs
+      - name: Measure coverage
+        id: coverage
+        if: github.event_name == 'pull_request'
+        continue-on-error: true
+        # Third-party action holding a write token, so pinned for supply chain security. v2.3.1, 2024-09-09.
```
schloerke, line 55: Thank you for comment

.github/workflows/build.yml
```diff
@@ -12,7 +12,80 @@ on:
...
+        uses: ArtiomTr/jest-coverage-report-action@7f750dd50f5585533321eb7ebc482b936b49a5d4
+        with:
+          annotations: coverage
+          # Generate report as markdown and then publish separately below
+          output: report-markdown
+
+      - name: Publish the coverage report
```
schloerke: What about `marocchino/sticky-pull-request-comment@v3` ?

Used in https://github.com/posit-dev/py-shiny/blob/e793b6c83accfe480f25deca8baf1290c397fca6/.github/workflows/verify-conventional-commits.yaml#L29C15-L29C56

.github/workflows/build.yml
```diff
@@ -12,7 +12,80 @@ on:
...
+          PR: ${{ github.event.pull_request.number }}
+          REPORT: ${{ steps.coverage.outputs.report }}
+        run: |
+          set -euo pipefail
+
+          # Always readable from the run's summary page, token or no token.
+          printf '%s\n' "$REPORT" >> "$GITHUB_STEP_SUMMARY"
+
+          # Then update the PR comment in place, so a busy PR gets one comment
+          # rather than one per push. Fails harmlessly on fork PRs.
```
schloerke: Move fork logic to `if` above. 

Silent failures are hard to diagnose. Make the failure loud or making the skip obvious

schloerke (approved):
I did not validate the tests. I trust they are correct given they are passing given minimal code adjustments.

.github/workflows/build.yml
jat255: I put the logic into a job-level `env.REPORT_COVERAGE` variable, which gaurds all three coverage steps

.github/workflows/build.yml
jat255: Good idea. I changed the action to use this action, and pinned it to v3.0.5 to match the convention we used in the other third-party action.

.github/workflows/build.yml
jat255: Switched to use `lts`

PR #247, opened 2026-08-11
test: Get the orphaned Playwright specs running again, as pytest

### LINKED ISSUE #243: Get orphaned Playwright specs running again and setup infrastructure for future tests

jat255:
Closes #243.

`playwright/` held four specs that nothing ran. All four are now pytest tests in `tests/`, next to the example suite from #238, and they run in CI.

**Stacked on #246** and rebased onto it, since both touch `Makefile`, `package.json`, `build.yml` and the README. Review that one first; GitHub retargets this to `main` when it merges.

## Groundwork

[`tests/conftest.py`](https://github.com/posit-dev/shinylive/pull/247/files#diff-e52e4ddd58b7ef887ab03c04116e676f6280b824ab7469d5d3080e5cba4f2128) — `fail_on_example_errors` becomes `fail_on_page_errors` and takes an opt-out, so a test that provokes an error on purpose marks itself `@pytest.mark.allow_page_errors` and asserts on the failure instead of being failed by the gate. #240 needs this for its syntax-error and download-failure tests.

Two things the review caught here, both worth knowing:

- **The gate would have failed on CI, not locally.** `build.yml` is the only workflow setting `GOOGLE_TAG_MANAGER_ID`, and a tag manager that fails to load is a console error. An autouse fixture now blocks analytics requests.
- **`--shard` split the whole collection, not the selected engine.** Conftest hooks run before `_pytest.mark`'s, so sharding happened ahead of the marker filter — balanced today only by luck. Moved to a `trylast` plugin; selection is unchanged (54 `py`, 22 `r`).

[`pytest.ini`](https://github.com/posit-dev/shinylive/pull/247/files#diff-fb6a686182f16eb54af3c628f38593f347f68aba31de903983023c560288d7a1) — every test carries an `examples` or `site` marker, enforced at collection. The two Make targets select on those markers, so a file with neither would run in *no* CI job; collection now refuses it rather than letting it disappear.

## The export fixture

[`tests/export_app.py`](https://github.com/posit-dev/shinylive/pull/247/files#diff-ee69b445eec6654069562c111198039628789386ed6a2a03b2fde5563192fce3) assembles a static export from what `make all` already produces under `build/`: `app.json`, `export_template/index.html` with its variables filled in, `edit/index.html`, and the bundle, symlinked rather than copied. That is everything `runExportedApp()` reads, so **nothing in this repo needs `pip install shinylive` any more** and the stale rationale comment in `build.yml` goes with it.

## The migrated tests

15 tests, from the four specs: [editor](https://github.com/posit-dev/shinylive/pull/247/files#diff-25b3894b44109986f03ff27a99d76821cd6a8b57e1161dca5eb7d8c3a9508d4c) and [URL loading](https://github.com/posit-dev/shinylive/pull/247/files#diff-ab8146ae7ce9202a64f9e4d7320c751b44e2f91f4b3f5f7ced7f0276e6bbcaf6) against the `_shinylive/` build the existing `static_server` fixture serves, and [static export](https://github.com/posit-dev/shinylive/pull/247/files#diff-ed38935bcecfeb6b986b211e047cd08521d5988477c86e66166852d64c864dc8) against the fixture above. App mode comes from `?_shinylive-mode=`, replacing a `sed` in [`playwright.config.ts`](https://github.com/posit-dev/shinylive/pull/247/files#diff-f679bf1e58e8dddfc6cff0fa37c8e755c8d2cfc9e6b5dc5520a5800beba59a92) that had not matched anything in `export_template/index.html` in a long time.

Three defects from the originals, are fixed in this PR:

- `expect_terminal_has_text()` returned a boolean that every call site discarded, so both terminal assertions in `examples-viewer.spec.ts` passed vacuously. The replacement asserts.
- "App view can show header bar" navigated to `/editor/`, so it had never tested the app view.
- `Mod-Enter` is Cmd on a Mac; the old spec pressed both modifiers to paper over it. `MOD_KEY` picks by platform, and both branches were checked against the real editor.

## Removals and CI

Gone: `playwright.config.ts`, the four specs and their fixtures, the `cypress:open` script, the `@playwright/test` dependency, the broken `test` target, and the commented-out playwright job. `examples-test-deps` becomes `test-deps`, since both suites use it.

The site tests run in [`build.yml`](https://github.com/posit-dev/shinylive/pull/247/files#diff-5c3fa597431eda03ac3339ae6bf7f05e1a50d6fc7333679ec38e21b337cb6721) right after `make all` — that build is what they test, and it reuses one that already takes most of the job. Two consequences worth stating explicitly: 
  - **A flaky site test now blocks the S3 deploy and the release upload**, not just the merge
  - `test-deps` plus ~55s of tests adds a few minutes to a job that currently runs in 2–4. The alternative is a seventh leg on `test-apps.yml`, which costs a second full `make all`.

## Verification

All 15 site tests pass against a local build, over three runs. The error gate was checked in all four directions — catches console errors, catches page errors, `allow_page_errors` skips both, clean pages pass — and the Google Tag Manager failure was reproduced against a real templated page and confirmed fixed. The examples suite still collects 91 and passes on a sample of both engines; `git diff` shows the Python side is byte-identical to what was reviewed before the rebase.

## Before merging

Revert `a177893`, which temporarily adds `242-jest-harness` to `build.yml`'s `pull_request` branches so this workflow runs while the PR is stacked. It becomes dead weight once #246 merges.

## Known gaps

Gate violations surface as pytest teardown ERRORs rather than FAILs — pre-existing from #238, and the exit code is still non-zero. The site suite covers the Python site only, as the originals did.


.github/workflows/build.yml
```diff
@@ -4,7 +4,11 @@ on:
   push:
     branches: [main, deploy]
   pull_request:
-    branches: [main]
+    # TEMPORARY: 242-jest-harness is here only so this workflow runs while #247
+    # is stacked on #246. It is what runs the site tests, so without it they go
+    # unverified. REVERT TO `[main]` BEFORE MERGING -- once #246 lands, GitHub
+    # retargets this PR to main and the entry stops doing anything anyway.
+    branches: [main, 242-jest-harness]
```
schloerke: TODO

README.md
```diff
@@ -152,18 +154,17 @@ Read the number for its direction, not its size. Coverage is collected only for
...
-Every app in `examples/`, loaded in a real browser against the built `_shinylive/` output, checked for tracebacks, warnings, output errors and console errors -- and, in the intent tests, driven through its inputs.
+Real browsers against the built `_shinylive/` output. Two halves, split by marker: the `examples` tests load every app in `examples/` and check for tracebacks, warnings, output errors and console errors -- driving each through its inputs, in the intent tests -- while the `site` tests cover the editor, apps loaded from a URL, and static exports.
 
 ```bash
-make examples-test-deps    # once
+make test-deps             # once
 make all                   # the tests drive the built output
-make examples-smoke-test   # both suites
+make examples-smoke-test   # every example app
 make examples-intent-test  # only the intent tests
```
schloerke: If there's a better name than `examples-smoke-test`. Maybe `examples-test-smoke` or `test-examples-smoke`?

I like to have prefix matching when possible to easier autocomplete

schloerke: Similar for `site-test` or `test-site`. ... idk which given the file

tests/conftest.py
```diff
@@ -147,6 +228,61 @@ def static_server() -> Iterator[None]:
             process.kill()
 
 
+@pytest.fixture(scope="session")
+def export_root(tmp_path_factory: pytest.TempPathFactory) -> Path:
+    """The directory the session's static exports are assembled in."""
+    return tmp_path_factory.mktemp("exports")
```
schloerke, line 234: Auto cleanup. Thank you

schloerke (approved):
Thank you!

schloerke:
Screenshot 2026-08-11 at 9 55 59 PM


Ooooooo. This is using Stacks. 

@jat255 (when you're ready to merge...) Can we click "Squash and merge stack 2" here to see if it makes to squash commits or just a single mega commit? Thank you

README.md
jat255: Went with `test-examples-smoke` and `test-site`, and gave the rest of the family the same prefix: `test-unit`, `test-unit-coverage`, `test-deps`, `test-examples-intent`. `type-check` and `examples-check-index` keep their existing names since they're checks, not full test suites.

I also changed workflow job ids and artifact names to follow the same pattern as the `make` targets. That also renames the status checks and I verified that `main` has no branch protection, no rulesets and no required checks that would be impacted by this change.

07b3228

jat255:
Final checks passed before reverting the temporary action change, 

image

So reverting, then merging

.github/workflows/build.yml
jat255: Removed immediately prior to merge

